Issue #256, opened 2015-02-13
Media file exceeding max file size present in the serialized form-model xml


ajaydivakaran:
Steps to reproduce:
1) Select a media file whose file-size exceeds the default 5mb.
2) Submit the submission.
3) Observe the xml sent to the server.

Actual: It contains the file-name of the media-file.
Expected: The file-name should not be present since it has violated the file size constraint.

Code observation:
Currently their is no check on file-size before updating the form-model.

Fix done on the forked repository:
Check for file-size before updating the form-model.

Would you be interested in a pull-request?

MartijnR:
Thanks!

Yes, PR very welcome.

PR #257, opened 2015-02-13
Pull request for github issue #256

ajaydivakaran:
Bugfix - Do not include file names the violate size constraint to form-model.

Currently the max file size logic is present in multiple places. In this commit I have not refactored that duplication.


MartijnR:
#256

MartijnR:
Went with the other solution instead 39441e42b57f761980cb2ea2e23df5ff4b32430a. I preferred not to duplicate the size check. May revert to your solution when redesigning the entire engine.

Let me know if the issue persists, please.

Thanks! :+1:

PR #268, opened 2015-04-16
Update print.js

maupanz:
Commenting this istruction, it's possible to use a new theme cloned from 'theme-grid'


MartijnR:
I have added this fix now.

PR #275, opened 2015-05-01
renamed $dataNodes to expected dataNodes

### LINKED ISSUE #274: Uncaught ReferenceError: $dataNodes is not defined
This shows up on `webform-edit-combined.min.js:10` when looking at console on chrome.

I think it has to do with [Form.js#L1388](https://github.com/enketo/enketo-core/blob/master/src/js/Form.js#L1388) and [Form.js#L1376](https://github.com/enketo/enketo-core/blob/master/src/js/Form.js#L1376) where `$dataNodes` was renamed to `dataNodes` - notice the missing `$`..

ukanga:
resolves #274


ukanga:
should have been a one line change but my vim settings tends to remove trailing spaces, should be ok though.

Issue #286, opened 2015-08-10
Remove dependency on require.js


alxndrsn:
Hi Martijn, not sure how feasible this is but it would be handy on client-side if `enketo-core` didn't require `requirejs`.  I'll probably take a look at this in the next few days, but would be interested to hear your thoughts on making the `requirejs` dependency optional, and if you'd accept a patch for that.

MartijnR:
I totally agree. I'd like to use CommonJS modules instead (probably with browserify). Backwards compatibility with requirejs would be nice-to-have (and I'm sure a lot of people would be very happy), but in my opinion it is not required if this gets too complex. If it breaks AMD-compatibility, we'd just make this v4.0.0.

So yes, I'd very much welcome a PR/patch for this!

alxndrsn:
Any thoughts on how widget config should be fetched, if not using `requirejs`?  It's the first non-obvious thing I've come to:
```javascript
//add widget configuration to config object
for ( var i = 0; i < config.widgets.length; i++ ) {
    id = 'text!' + config.widgets[ i ].substr( 0, config.widgets[ i ].lastIndexOf( '/' ) + 1 ) + 'config.json';
    widgetConfigFiles.push( id );
}

//load widget config files
require( widgetConfigFiles, function() {

    for ( var i = 0; i < arguments.length; i++ ) {
        widget = JSON.parse( arguments[ i ] );
        widget.path = config.widgets[ i ];
        widgets.push( widget );
    }

    deferred.resolve( widgets );
} );
```
> https://github.com/enketo/enketo-core/blob/master/src/js/widgets.js#L104

Are users expected to change widget configs?

And is there anything that's needed pre-loadtime apart from `options.selector`?  If not, would it work to convert enketo's `config.json` `widgets` object from `["widget-1", ... , "widget-N"]` to be a map `{"widget-1":".1-selector", ..., "widget-N": "N-selector"}`?

E.g for https://github.com/enketo/enketo-core/blob/master/config.json#L3
```json
{
    "widgets": [
        "enketo-widget/note/notewidget",
        ...
```
becomes 
```json
{
    "widgets": {
        "enketo-widget/note/notewidget": ".note",
        ...
```
I think this would remove the need for accessing individual widgets' `config.json` files from the JS.

MartijnR:
[...]

PR #287, opened 2015-08-14
Convert to common JS

alxndrsn:
Thought you might be interested to see progress.  Bundling with browserify seems to be working nicely.

This isn't ready to be merged yet because:
* the tests are broken (these still expect requirejs)
* requirejs is not fully working.  I *think* this may be as simple as getting the config loading to accept either JSON or javascript objects
* `Makefile` should probably be converted to `Gruntfile`
* `patch` script shouldn't have been committed

Also I've turned off beautifier when running `grunt test` so that the PR diff is cleaner.


alxndrsn:
Beginnings of a fix for #286

MartijnR:
Thanks! Will look at this with the greatest interest! (on Monday)

Makefile
```diff
@@ -0,0 +1,65 @@
...
+compare-built:
+	ls -al build/js/*.js
+
+.PHONY: compile-dependencies
+compile-dependencies:
+	mkdir -p build/fetch/cc
+	(cd build/fetch/cc && \
+		wget -c http://dl.google.com/closure-compiler/${CC_VERSION} && \
+		tar -xf ${CC_VERSION} && \
+		cp compiler.jar ../../lib)
```
MartijnR: mkdir build/lib

MartijnR:
Very cool you got this working so quickly! Some questions, comments:

* yes, grunt (or gulp) preferred
* closer compiler is probably run in SIMPLE_OPTIMIZATIONS mode (because I gave up on ADVANCED_OPTIMIZATIONS long time ago). In that case, I'd prefer to stick with a uglify2 in order not require java on the production (and dev) server. 
* Probably best to rename app.browserify.js to app.js
* Seems feasible to get rid of browserify-shim in future, right? Just required for xpath?

src/widget/geo/geopicker.js
```diff
@@ -1397,16 +1405,20 @@ define( [ 'jquery', 'enketo-js/Widget', 'text!enketo-config', 'leaflet', 'q' ],
         $.fn[ pluginName ] = function( options, event ) {
```
MartijnR: Deliberately changed boilerplate only in Geopicker?

require-config.js
```diff
@@ -1,21 +1,20 @@
 require.config( {
```
MartijnR: If this file is only used in (the old) app.js, I guess we should get rid of entirely.

src/widget/geo/geopicker.js
alxndrsn: Didn't realise this was boilerplate.  Was fiddly getting this working, and difficult to work out what was breaking, hence all the extra logging.  Happy to revert before merging, but would prefer to keep it in until then...

alxndrsn:
> * yes, grunt (or gulp) preferred

Will do.

> * closer compiler is probably run in SIMPLE_OPTIMIZATIONS mode (because I gave up on ADVANCED_OPTIMIZATIONS-friendly JSDoc commenting long time ago). In that case, I'd prefer to stick with a uglify2 in order not require java on the production (and dev) server. Does that make sense, or is there a good argument to switch to using Closure Compiler?

Will switch to uglify2.

> * Probably best to rename app.browserify.js to app.js

Will do.

> * Seems feasible to get rid of browserify-shim in future, right? Just required for xpath?

Yes, I think so.  I seem to have shimmed jQuery too, but I'm fairly sure that's unnecessary.

alxndrsn:
```
TODO
[ ] convert Makefile to grunt
[ ] update README
[ ] fix tests
[ ] remove grunt-contrib-require-js
[ ] remove requirejs from package.json
[ ] remove requirejs from bower.json
[ ] turn on beautifier and beautify everything
[ ] fix uglification step in `Gruntfile`

**Comment by alxndrsn** [comment]
Some updates:
* rebased onto current master
* fixed tests to work with CommonJS modules
* converted `Makefile` to `grunt`
* add uglification step to Gruntfile
* remove all references to requirejs (except for AMD module definitions in `src/js/**/*.js`)

Some stuff that still needs work:
* when running tests, there are lots of warnings about missing widget config "modules"
* readme will need updating to reflect new usage
* beautification is still disabled, as it makes it much harder to compare pull requests and handle rebasing (best to wait until this branch is in a mergable state before beautifying IMO)

**Comment by MartijnR** [comment]
This is so cool. Thanks for this first major contribution (by anyone)! I see the widget loading error was [already existing](https://travis-ci.org/enketo/enketo-core). I'll look at that.

Will also look at the filler issue. Would seem to signify that the horizontal-choices widget is not loaded.

Am going to see if enketo-express will just run using your enketo-core branch. Because of the backwards-requirejs-compatibility it should still work, I think.

**Comment by MartijnR** [comment]
Just gathering some stuff here (I will be updating this comment), just in case you're working on this branch still to avoid conflicts:

The horizontal choice widget contained an error (in master). [This line](https://github.com/enketo/enketo-core/blob/master/src/widget/horizontal-choices/horizontalchoices.js#L25) should be: 
```
        Widget.call( this, element, options );
```

grunt.watch.js.task = ['browserify']

**Comment by alxndrsn** [comment]
Glad you like it!  I'll check here again tomorrow and see if I need to follow up on anything :+1:

**Comment by MartijnR** [comment]
Am somewhat stuck with the dynamic widgets loading in requirejs. I think we can simplify the widgets system by letting each widget return `{name: widgetName, selector: theselector }` and getting rid of all the widgets' config.json files. We'd then have to force a strict naming convention for the scss files.

**Comment by alxndrsn** [comment]
Tricky indeed.  I've been using [`almond`](https://github.com/jrburke/almond), so loading was synchronous even with requirejs syntax.

Would it work to have an extra adapter for requirejs users to preload the configs and widget JS?  Or would this significant impact the user experience?

Otherwise, simplest approach may be to branch `_getWidgetConfigs()` to use asynchronous require when AMD is detected - shouldn't be a massive amount of different code, and we have both implementations available in `master` and `commonjs`

**Comment by MartijnR** [comment]
I'd rather not do amd specific stuff. I think the widgets flexibility is just over-engineered and that nobody is actually using that flexibility. I've asked the question [here](https://groups.google.com/forum/#!topic/enketo-dev/V09AyrDg3AM) to get feedback. In the mean time, I'll just move ahead with simplifying the loading by getting rid of all the widgets' config.json files.

**Comment by MartijnR** [comment]
I wonder if we can create a module that just requires each widget and exports an array of these widgets.

```
module.exports = [
    require('./path/to/widgetA'),
    require('./path/to/widgetB',
    etc.
];
```

This file can then perhaps be overwritten using an alias in browserify or requirejs config. That alias could be a optional configuration setting to bring widget loading flexibility back.

MartijnR:
Alex, I changed some stuff. If you have time, please let me know what you think. Otherwise, no problem either.

Haven't gotten the alias to work as a way to load an alternative widgets module, but it should be possible.

MartijnR:
The flexibility is back now, by overwriting 'browers.widgets' in package.json

MartijnR:
Merged in 4.0.0. Thanks again!

PR #294, opened 2015-08-31
Fix initialisation issue on Android

alxndrsn:
This fixes a bug on Android described in https://github.com/medic/medic-webapp/issues/1207


MartijnR:
Thanks! This was definitely not according to the spec!

PR #295, opened 2015-08-31
Fix incorrect key matching in selectpicker

alxndrsn:
From @browndav as per https://github.com/medic/medic-webapp/commit/8221362f27a505778396b5f6a9c1e28b4c91c28c


MartijnR:
Makes sense! Thanks. (Would like to get rid of this selectpicker in future).

PR #297, opened 2015-09-01
Fix typo in desktop/selectpicker definition

alxndrsn:
I *think* this is a typo; it breaks widget loading on my machine.

### DIFF
--- a/src/widget/select-desktop/selectpicker.js
+++ b/src/widget/select-desktop/selectpicker.js
@@ -244,7 +244,7 @@ define( function( require, exports, module ) {
         } );
     };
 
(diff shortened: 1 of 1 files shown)


alxndrsn:
Actually this fix is incorrect.  Wrapping in parens seems to be 'better', i.e.:

```
(function( $ ) {
}( jQuery ));
```
However, I'm not yet sure the intent of the code.

alxndrsn:
(PR updated to wrap function in parens)

alxndrsn:
I seem to be confusing this with something else - the current code works fine, although I don't understand the syntax(!)

PR #300, opened 2015-09-01
Faster travis builds?

alxndrsn:
[Apparently][1], explicitly disabling `sudo` on the Travis build should speed things up.

[1]: http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade


MartijnR:
very cool. let's see

PR #305, opened 2015-09-03
Add missing `var` delcaration for `widgets`

alxndrsn:


MartijnR:
Thanks! So cool to have second pair of eyes!

PR #308, opened 2015-09-09
Don't include `config.json` in the SASS watch-list

alxndrsn:
Pretty sure it's no longer necessary to recompile SASS if `config.json` changes.


MartijnR:
:+1:

PR #312, opened 2015-09-16
Change logging level to `log` if errors are deliberately ignored

alxndrsn:
Just a suggestion - these messages come up in scary red writing in my console, even though the message is to be ignored.  Seems more appropriate to me to make these plain `console.log` messages.


MartijnR:
Agreed. Thanks!

Issue #311, opened 2015-09-16
Custom validation for custom data types


alxndrsn:
Following on from #310, is it possible to have custom validation for a custom data type?  I was hoping this could be added using a widget, but looking at the code around https://github.com/enketo/enketo-core/blob/master/src/js/FormModel.js#L1033, that doesn't seem to be the case.

My current use case is validating a phone number field.  I'm not sure if that could be done with XPath, and ideally I'd use [libphonenumber](https://github.com/googlei18n/libphonenumber) in the browser.  My current workaround is to use my widget to hide the real `<input>`, and only copy the value to that from a fake `<input>` when my custom (JS) validation passes.

MartijnR:
You can actually use the regular constraint validation (with the regex() function) for this. That would be the preferred solution I think (and would work in any OpenRosa-compliant client).

I haven't thought about adding the ability to extend FormModel with new types. Moving `types` to a separate module and then optionally overriding this entire module might be a decent solution.

alxndrsn:
Sadly it's not very regexable, but I'll have a look at modifying `types` somehow...

alxndrsn:
Turns out this would be quite a simple change.  I'm not sure if that commit/PR fits with the code style, but exposing the `types` object from `FormModel` allows very simple injection of validators for custom types: https://github.com/alxndrsn/medic-webapp/blob/647c27dbced878be32cd9ee8a2bf33e92c3726d8/static/js/enketo/widgets/phone-widget.js#L17

alxndrsn:
I think this issue was resolved by c0efdac :smile:

PR #313, opened 2015-09-17
Make `FormModel.types` accessible publicly

alxndrsn:
This allows widgets or other extensions to add validation for custom data types.

Issue: #311

### DIFF
--- a/src/js/FormModel.js
+++ b/src/js/FormModel.js
@@ -1188,5 +1188,7 @@ define( function( require, exports, module ) {
         }
     };
 
+    FormModel.types = types;
+
     module.exports = FormModel;
 } );


MartijnR:
Of course! I like it! I may want re-arrange code, but will look later.

alxndrsn:
:+1:

MartijnR:
Alex if this becomes a prototype property, it will still work for you, right?

alxndrsn:
Would I just change this line:
https://github.com/alxndrsn/medic-webapp/blob/901e858912365962a0a5f419a627826908b6cd65/static/js/enketo/widgets/phone-widget.js#L19
to read
```javascript
FormModel.prototype.types.phone = ...
```
or would I need to apply the type changes to each instance of `FormModel` via e.g. `myForm.getModel().types`?

MartijnR:
Yes, I think so too. This should better as you won't have to changes the instances. I'll merge this now.

alxndrsn:
Great, thanks!

MartijnR:
thank you

PR #318, opened 2015-09-30
Use `prop(..., false)` instead of `removeAttr(...)` in filepicker

alxndrsn:
* `disabled` should be controlled via jQuery's `prop()` method rather than
   attr methods
* `removeProp()` is not suitable, because it prevents the field ever becoming
  disabled again:

  > Do not use this method to remove native properties such as checked, disabled, or selected. This will remove the property completely and, once removed, cannot be added again to element.
  > — _https://api.jquery.com/removeProp/_


MartijnR:
Thanks. Merged!

PR #319, opened 2015-09-30
Toggle button states using `disabled` attribute instead of class

alxndrsn:


MartijnR:
This seems to break the whole button enabling/disabling in pages mode and non-pages mode. I tried to fix but came across a situation where the property `disabled` is false and the disabled attribute was present. Then I got tired and started wondering if this is worth the effort...

alxndrsn:
Aagh, apologies!  Was the problem reproducible in the `index.html` demo page?

I will give it another go in the future.

MartijnR:
No worries. Yes, it was reproducable in index.html.

PR #321, opened 2015-09-30
Remove unused vars

alxndrsn:
Ran jshint to see if there were unused vars.
I considered turning on the `unused` option in `.jshintrc`, but there are a few vars in method signatures that I assume are wanted kept just-in-case.


MartijnR:
Great!  Merged. Thanks.

Issue #320, opened 2015-09-30
Change validation to use promises


alxndrsn:
If field and form validation could be converted to use `Promise`s, it would make async validation possible on (custom) fields.

MartijnR:
Not sure. Is there a use case for async validation _that would work offline_?

alxndrsn:
A very reasonable question!

Our main use case is that we want to check certain fields for uniqueness against our DB before saving the form data.  We're using [PouchDB]() for storage, and all its APIs are asynchronous.

I think the same case would apply if using IndexDB - according to MDN, [_operations performed using IndexedDB are done asynchronously_](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API#Synchronous_and_asynchronous).

I get that this is probably not a priority for Enketo.  However, the changes weren't extensive so if this isn't something you want on the mainline, then it shouldn't be too much trouble maintaining a fork, or working out a neat way of adding an async `Form.prototype.validate`.

MartijnR:
Ah yes, thanks. I think that's a good use case especially for hybrid apps that sync data two-ways. PR welcome (I use Q.Promise elsewhere until there is cross-browser promise support) . I hope I'm not overlooking a potential issue in enketo-express. As far as I remember, there are 3 validation scenarios:

alxndrsn:
:+1: thanks

PR #323, opened 2015-10-07
Convert form validation to use Promises.

alxndrsn:
Issue: #320


MartijnR:
Looks nice! (except that it reminds me what a mess I made with the validateInput method... - one step at a time)

Can I add you as a contributor in package.json? If that's okay, please let me know if you'd like me to add the email address you are using in your commits (or none, or another).

alxndrsn:
Thanks, that would be very nice.  Please don't list an email address.

Issue #331, opened 2015-10-14
Form does not scroll to top on page flip if it's in side a scrolling container


alxndrsn:
# Observed

Automatic scroll to the top of the page on page-flip does not work if the form is enclosed in a scrolling region.

# To recreate

1. Modify `forms/index.html` as follows:
```html
<style>
    ...
        article {
                width: 300px;
                height: 200px;
        }
        form {
                overflow: scroll;
        }
</style>
```

2. `grunt server`
3. open your browser at http://localhost:8005/forms/?xform=https://raw.githubusercontent.com/alxndrsn/enketo-client-side-transformer/master/xforms/pregnancy.xml
4. browse to second page
5. scroll to bottom of form
6. click next
7 observe that you only see the bottom of the next question

# Fix

I think `window.scrollTo` is the culprit here - obviously useful if you want to scroll to the top of the **window**, but something like `pageEl.scrollIntoView()` might work better.

MartijnR:
Great. I wasn't even aware of the existence of scrollIntoView!

PR #332, opened 2015-10-14
Scroll to the top of the form when flipping pages

alxndrsn:
If the scrolling region for the form is not actually `window`, then the old code
would not scroll to the top of it.

This commit replaces use of `window.scrollTo()` with
`$( formSelector ).scrollIntoView()`.

Issue: #331


MartijnR:
:+1:

Issue #326, opened 2015-10-09
Check current page for errors before moving to Next page


alxndrsn:
As suggested in #320:

> upon page next click, block UI to check current page for errors

I don't think it will be necessary to check on Prev.

alxndrsn:
Is there any chance of problems if earlier fields depend on input in later fields to pass validation?  Seems like a very weird/bad/incorrect form design, but perhaps others have seen valid uses?

MartijnR:
Thanks for adding this issue.

Yes, there is definitely an issue with that. However, this would make it behave closer to ODK Collect. There have been some requests to do this (https://github.com/kobotoolbox/enketo-express/issues/300).

I guess this would be a case of the benefits outweighing the 'costs'.

If it is not necessary to have such forward-looking logic, I guess only the default (non-pages) mode would be an option. However, we could also consider adding a setting (a class on the body) that does not add the page-flip-validation.

abbyad:
From my experience you encounter the same form design issue in Collect already. If you have validation that relies on a later field the validation for the earlier page will always fail. 

If you find yourself in this situation you can skip directly to another field in Collect, but not in Enketo.

MartijnR:
Yes, with KoBo we were discussing to add some way to bypass like ODK Collect. But it seems to me that is a bit weird. Why have the protection and allow bypassing it? Not sure, including whether to still keep 'Go to End'.

abbyad:
Skipping from section to section may make sense for some long forms (something like big government tax forms), however we've never had a use for it yet. For what its worth, we are hiding 'Go to End' right now.

MartijnR:
Thanks. So you just have Back, Next and Go to Beginning. I think that makes sense when the flow is restricted. 

So maybe this:
page-validate enabled (default): Back, Next, Beginning
page-validate disabled: Back, Next, Beginning, End (maybe in future: Jump to section)

abbyad:
We removed Go-to-Beginning for now pending more user testing, but the way you have it makes sense for general use.

alxndrsn:
I think this can be closed.

MartijnR:
Yes, thanks for that. Very useful addition. I asked some people to comment on the new feature, wrt to the First/Last buttons and a possible Jump-to-section. I am leaving it open for a weeks to see if they want to comment.

ccollins0601:
Hi Martijn,
This is great. We're trying to break it now. So far so good, will let you know if we find any problems.

[...]

PR #333, opened 2015-10-14
Check current page for errors before moving to Next page

alxndrsn:
Issue: #326

### DIFF
--- a/src/js/Form.js
+++ b/src/js/Form.js
@@ -388,13 +388,20 @@ define( function( require, exports, module ) {
             },
             next: function() {
                 var next, currentIndex;
+                var that = this;
                 this.updateAllActive();
                 currentIndex = this.getCurrentIndex();
-                next = this.getNext( currentIndex );
 
-                if ( next ) {
-                    this.flipTo( next, currentIndex + 1 );
-                }
+                form.validateContent( this.$current )
+                    .then( function( valid ) {
+                        if ( !valid ) return;
+
+                        next = that.getNext( currentIndex );
+
+                        if ( next ) {
+                            that.flipTo( next, currentIndex + 1 );
+                        }
+                    } );
             },
             prev: function() {
                 var prev, currentIndex;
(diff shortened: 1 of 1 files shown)


src/js/Form.js
```diff
@@ -1986,9 +2004,11 @@ define( function( require, exports, module ) {
                 if ( eventType === 'validate' ) {
                     that.setInvalid( $input, 'required' );
                 }
+
+                valid = Promise.resolve( false );
```
alxndrsn: This fixes a (potential) bug in the Promise validation code - if the only validation error was a `required` field, it is otherwise ignored.

src/js/Form.js
```diff
@@ -1834,19 +1841,27 @@ define( function( require, exports, module ) {
...
          */
         FormView.prototype.validateAll = function() {
+            return this.validateContent( $form );
+        };
+
+        /**
+         * Validates all enabled input fields in the supplied container, after first resetting everything as valid.
+         * @return {Promise} wrapping {boolean} whether the container contains any errors
+         */
+        FormView.prototype.validateContent = function( container ) {
```
alxndrsn: As `container` is now either the `<form>`, or a `.page`, and because the `.page` could also be a `.question`, we have to `addBack()` quite a few times.

This is a little frustrating, and it's tempting to monkeypatch `jQuery` with something like:
```js
$.fn.findIncludingSelf = function(selector) {
    return this.find(selector).addBack(selector);
}
```

src/js/Form.js
```diff
@@ -388,13 +388,20 @@ define( function( require, exports, module ) {
...
+                form.validateContent( this.$current )
+                    .then( function( valid ) {
+                        if ( !valid ) return;
+
+                        next = that.getNext( currentIndex );
+
+                        if ( next ) {
+                            that.flipTo( next, currentIndex + 1 );
+                        }
+                    } );
```
alxndrsn: This bit ^^ is the new functionality - everything else is just refactoring and fiddling.

src/js/Form.js
MartijnR: Yes, indeed. Seems like I should have had a test for that...

MartijnR:
Great stuff. Thanks a lot!

I squashed them into one commit and since I was manipulating anyway I also did this:
- changed commit message (this is a bit anal, but I try to start each message with either _changed: , added: , fixed: , or removed:_  which corresponds with the changelog sections. The idea is that we could auto-generate a change log in the future.
- changed container to $container for consistency

How does it affect your project if I squash, change PR commits before merging? If it's a pain, I won't do it any more. I could either just merge or ask you for changes.

PR #335, opened 2015-10-15
Move event handling/retrigger out of `FormView.validateInput()`

alxndrsn:
Just a minor change that I noticed when rebasing the `medic-mobile` fork - seems more logical to move event propagation to the event handler, as described below.

-----

Move the triggering of `'valuechange.enketo'` out of `FormView.validateInput()`, and into the event handler for `'change.file validate'`.

* this is the only place `eventType` can have a value of `'change'`
* this makes the behaviour of `validateInput()` more discrete


MartijnR:
Yes, indeed. Thanks!

PR #345, opened 2015-11-19
Only create new reqSubtle element if it's required

alxndrsn:


alxndrsn:
Travis failed because `npm` is down.

MartijnR:
Thanks. Looks better! 

Bootstrap-timepicker has dissappeared from npm: https://github.com/kobotoolbox/enketo-express/issues/359

PR #348, opened 2015-12-02
Remove no-op assignments from widgets

alxndrsn:
Looks like `data` is not used again after it was reassigned.


MartijnR:
Yes, looks like you're right! Is the case in all widgets it seems. Do you want to add the others too perhaps (or should I get off my lazy ass and do this - you can tell me)?

alxndrsn:
I'll take a look :)

alxndrsn:
Updated.

MartijnR:
Thanks!

Issue #356, opened 2016-02-01
Adding repeated item immediately scrolls window to top of page


alxndrsn:
It seems like when a repeated item is added, the window immediately scrolls to the top.  This has no effect if the scroll is already at the top, but if the user has scrolled down a bit, then it can be disorientating.

Ref. medic/medic-webapp#1588

MartijnR:
Thanks! Sounds like a bug. It shouldn't call focusOnFirstQuestion [if the page doesn't change](https://github.com/enketo/enketo-core/blob/master/src/js/Form.js#L445), but apparently it does.

alxndrsn:
Thanks.  I'll have a look at the proper fix at some point, but for now we're getting on fine with the example in the PR.

MartijnR:
I think this does the trick, hopefully without breaking stuff on mobile screens (the scrollIntoView() was only required for mobile if I remember correctly).

This avoids the double scrolling of the PR (first to top, then to element).

alxndrsn:
Thanks, will give this a whirl in the next couple of days/

alxndrsn:
:+1: works good - now no scrolling occurs at all

MartijnR:
Great. Thanks for checking!

PR #357, opened 2016-02-01
Scroll repeating element into view when it is added

alxndrsn:
Issues: #356 medic/medic-webapp#1588

I wrote this fix a while back, and on reviewing, I suspect the TODO could be removed - it's likely that the `scrollIntoView()` call is useful whether the new elements are on the same page or not.


MartijnR:
Thanks. As you already indicated in your comment, this should probably be fixed elsewhere. It looks like the flipTo function has a bug.

MartijnR:
used alternate fix: 71ef561d12a76d9e467dd3c2cc5d986c2f977d41

PR #371, opened 2016-05-24
Update Form-model.js

theywa:
adoptNode is not supported on IE 11 so need alternative function to prevent error and allow saved value show up on edit form

### DIFF
--- a/src/js/Form-model.js
+++ b/src/js/Form-model.js
@@ -153,6 +153,29 @@ define( function( require, exports, module ) {
     FormModel.prototype.node = function( selector, index, filter ) {
         return new Nodeset( selector, index, filter, this );
     };
+    
+    /**
+     * Alternative adoptNode on IE (http://stackoverflow.com/questions/1811116/ie-support-for-dom-importnode)
+     */
+    FormModel.prototype.importNode = function(node, allChildren) {
+        switch (node.nodeType) {
+            case document.ELEMENT_NODE:
+                var newNode = document.createElementNS(node.namespaceURI, node.nodeName);
+                if(node.attributes && node.attributes.length > 0)
+                for(var i = 0, il = node.attributes.length; i < il; i++)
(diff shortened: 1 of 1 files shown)


MartijnR:
Thanks! Are you sure that fixes the IE problems with loading an existing instance? I thought I tried this.

If it does: 

That is great! 👍 Would it work to check for the feature instead of the useragent? =>

```if (typeof this.xml.adoptNode === 'function')```

Running `grunt jsbeautifier:fix` should fix the code style issues.

MartijnR:
Isn't importNode supposed to be supported in IE11? In other words, do we need the polyfill?. Note, that we do not need (or even want) support for older versions of IE due to so many other problems.

theywa:
if your question "Are you sure that fixes the IE problems with loading an existing instance?" refer to issue on IE 11 edit form there will be no saved value show up ? If yes then yes it solved that issue.
Using "adoptNode" will result error "No such interface supported" also using "if (typeof this.xml.adoptNode === 'function')" will create that issue too.. 
here some demo http://supadosworks.com/videos/demo-issue/OC-7124/OC-7124-fix-show-value-on-edit.ogv.
Some article say importNode and adoptNode not supported in IE but I'm not sure if it including IE11,. Not tested on older version IE yet and I also don't know if we need polyfill or not.

MartijnR:
Thanks!

Nice of Microsoft... So document.adoptNode is a function, it just doesn't work...

Document.importNode is also a function. If you can check whether that one actually works in IE11 without the polyfill, that would be great.

screen shot 2016-05-25 at 8 01 29 am

theywa:
I tried that, but like adoptNode, document.importNode and this.xml.importNode also generate same error on IE. It's look like that polyfill needed at least until we found better solution,.

MartijnR:
Oh man... I can't wait for IE11 to die... (and then Safari).

Will test and then merge!

MartijnR:
👍  I just had to change a few remaining code style and jshint issues, so I added them, squashed, and merged.

Thanks!

Next time, it would be great if you could try to make `npm test` pass.

PR #396, opened 2016-12-22
Make the root URL of demo server serve the demo page

alxndrsn:
Not sure if there's a technical reason for the current config, but this seems more convenient for dev use - open the browser to the address mentioned below (`http://localhost:8005`), and you immediately get the form demo page:

```
Running "concurrent:develop" (concurrent) task
    Running "watch" task
    Waiting...
    Running "shell:transformer" (shell) task
    Running "connect:server:keepalive" (connect) task
    Waiting forever...
    Started connect web server on http://localhost:8005
    app running on port 8085!
```

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -27,7 +27,8 @@ module.exports = function( grunt ) {
         connect: {
             server: {
                 options: {
-                    port: 8005
+                    port: 8005,
+                    base: ['forms', 'build']
                 }
             },
             test: {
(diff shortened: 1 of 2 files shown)


alxndrsn:
@MartijnR This PR is deliberate, but I meant to push it to my own repo.  Do you mind where "experimental" branches like this go?

I'm checking the Travis failures now.

MartijnR:
Thanks! If there is a technical reason I no longer remember it. Will test, squash and merge!

This is perfect. I use fix/ and feature/ branches just like you did.

MartijnR:
I just found out why it was like it was, after committing. It was related to enketo-core-performance-monitor (which requires css file references in index.html to relative to its own location. I made a small change to facilitate this by moving index.html out of forms into build.

PR #404, opened 2017-03-14
Support Chrome's UNORDERED_NODE_ITERATOR_TYPE

alxndrsn:
Hi Martijn, just getting back on the wagon after a long time away from Enketo.  Here's a fix for a weird bug (either in enketo, or Chrome ;¬) which I'm probably the only person in the world to suffer from!  Please let me know if it's suitable for the enketo mainline, and if not if there is anything I can do to improve it.

-----

When chrome's XPath `evaluate()` method is asked for an `ORDERED_NODE_SNAPSHOT_TYPE`
it can return an empty `UNORDERED_NODE_ITERATOR_TYPE`.  This throws an error when
`snapshotLength` is called on 

```
TypeError: Failed to read the 'snapshotLength' property from 'XPathResult': The result type is not a snapshot.
```

On the other hand, there are other types which will throw a similar error if
`iterateNext()` is called on them.


MartijnR:
Welcome back! 🌟 

Did you mean to add a test with that XForm?

alxndrsn:
* Ahem * no!  Didn't mean to add the form at all (¬;

MartijnR:
Ok, will try to reproduce by creating a test that fails with grunt karma:browsers. So just a 'nodes' result that is empty, I presume?

alxndrsn:
> Ok, will try to reproduce by creating a test that fails with grunt karma:browsers.

I'm not sure that will work - the problem shouldn't appear when using XPathJS because the native evaluator won't be called: https://github.com/medic/enketo-core/commit/d889014f214fbde968b4229009d09ffe95e2ed14#diff-3647f2f20ebf195842b4f1664089ed79

Because of that commit, I [modified our `openrosa-xpath-evaluator`](https://github.com/medic/openrosa-xpath-evaluator/commit/195bf58fee59f8ba04f0c2940c32e32b6b88764b) to delegate requests for `*_NODE_*_TYPE` to the browser's evaluator, which caused the bug fixed above.  I hope this delegation is reasonable, at least barring bugs in Edge.

Hence my initial comment that _I'm probably the only person in the world to suffer from_ this bug.  Apologies for failing to provide more context earlier.

-----

I think I should look into this a little further though - on re-reading my code there's a chance I've caused this by changing the requested `resultType` for the `evaluate()` call from `ORDERED_NODE_SNAPSHOT_TYPE` to `ANY_TYPE`.  I'll re-open if I can't fix at this end!

MartijnR:
Ah thanks!

If you find it's an issue, we should definitely be able to write a test that forcefully evaluates with the native XPath evaluator. Enketo also sends native XPath expressions as much as possible to the native evaluator (in a much less intelligent way, if the parameter tryNative is `true` and no custom function calls are detected).

Issue #336, opened 2015-10-19
Provide method for obtaining Model without non-`relevant` fields


alxndrsn:
> #### 6.1.4 The relevant Property
> 
> Description: indicates whether the model item is currently _relevant_. Instance data nodes with this property evaluating to `false` are unavailable in the user interface and can be removed from submission serialization.
> — _http://www.w3.org/TR/xforms/#model-prop-relevant_

It would be very helpful to have a method similar to `Form.getDataStr()` which would allow fields with `relevant` evaluating `false` excluded.

Alternatively, `FormModel.getStr()` could be modified to always exclude these fields without breaking the spec.  However, this might have surprising effects if some users are relying on non-relevant data appearing in their form submission.

MartijnR:
You're right. This should actually be done. First we should find out if ODK Collect excludes those correctly (I think it does).

If it does, I'd be happy to get this added in a way that keeps the default behaviour the same (because it's definitely going to break stuff in various Enketo apps). Implementation will be messy as far as I can see. I think accessing form.getDataStr() with a non-default parameter that excludes relevants perhaps.

The reason I have never done this, is that the benefits of a smaller submission (or local storage size) didn't seem worth the effort for the current Enketo apps. It has become quite a bit more feasible to do this though with a recently-introduced better way of merging (partial) records into the model.

MartijnR:
confirmed that ODK Collect strips irrelevant fields from submission. They may actually simply remove all empty fields. [To check, because if so, then all servers would already support this]

That would be super-easy to implement too.

MartijnR:
depending on implementation this would also do b44dbd8bae1751f2741bb365d18fb558ff841976

MartijnR:
Don't strip relevant empty fields. It's helpful to know the distinction between empty vs. empty && irrelevant.

dorey:
Ping @eawag-rdm, who created eawag-rdm/kobo-post to focus on this issue.

I've started a discussion with and tested out how ODK Collect and Enketo preserve nodes of the XML submissions and I want to report that there might be another way to similarly flag fields that were skipped in Collect. Please see the results from submissions to this contrived example:

```
<!--
3 submissions from ODK Collect
--------------------------------
Minimized XML to demonstrate how collect strips nodes.
All comments are added by me. -@dorey
--> 

<!-- submission 1 . all questions have responses -->
<amJipsGQEQknybsUZp3qJw>
    <q1>Collect'd</q1>
[...]

PR #409, opened 2017-03-24
added: method to obtain primary instance without irrelevant nodes, cl…

MartijnR:
…oses #336

Since you created issue #336, I figured I'd try asking you to review it @alxndrsn :simple_smile:

### DIFF
--- a/src/js/Form.js
+++ b/src/js/Form.js
@@ -125,11 +125,17 @@ define( function( require, exports, module ) {
         };
 
         /**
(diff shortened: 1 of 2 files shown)


alxndrsn:
Looks good to me!

I do wonder if `getNodesToUpdate()` should be renamed now, as that doesn't seem to be its only use...

MartijnR:
Thanks! Yes, good point, maybe `getRelatedNodes` or `getReferredNodes` would be better

PR #410, opened 2017-03-27
Reference the translator module consistently from widgets and Form.js

alxndrsn:
Make references to `translator` consistent.

Ideally though, all code would probably refer to `fake-translator`, and leave this up to 3rd-parties to override with their own translator using e.g. browserify.

### DIFF
--- a/src/js/Form.js
+++ b/src/js/Form.js
@@ -26,7 +26,7 @@ define( function( require, exports, module ) {
     var $ = require( 'jquery' );
     var Promise = require( 'lie' );
     var utils = require( './utils' );
-    var t = require( './fake-translator' ).t;
+    var t = require( 'translator' ).t;
     var pkg = require( '../../package' );
     var config = require( 'text!enketo-config' );
     require( './plugins' );


MartijnR:
Thanks! Totally right about that error.

> Ideally though, all code would probably refer to fake-translator, and leave this up to 3rd-parties to override with their own translator using e.g. browserify.

This is exactly the idea behind require(`translator`) and having a dummy fake-translator included that doesn't translate anything. Why would calling `require('fake-translator')` be better?

alxndrsn:
> Why would calling `require('fake-translator')` be better?

There's [already a published node module called `translator`](https://www.npmjs.com/package/translator), which is similar but has a different API.

That library doesn't look hugely popular so it's unlikely to ever cause a problem, but I was thinking if the enketo code used a relative require (e.g. `require('./fake-translator')` then that might be overrideable but also be guaranteed collision-free.

This would be similar to what is done with `require( './xpath-evaluator-binding' );` in `Form-model.js`.

MartijnR:
Thanks. Yes, that's a good reason. The only problem is that it would break things in Enketo-Core consuming apps that are already overriding `translator`. Maybe good to keep for a 5.x version.

alxndrsn:
> Maybe good to keep for a 5.x version.

If it was changed for `translator`, then I guess `widgets` would also be worth changing.

MartijnR:
As I'm about to make this even worse with a `gui` module that can overwrite native alert() and confirm() dialogs, how about using something like `require('enketo/translator')` with this aliasify config:

```
"enketo/config": "./config.json",
"enketo/widgets": "./src/js/widgets.js",
"enketo/translator": "./src/js/fake-translator",
"enketo/gui": "./src/js/fake-gui"
```
We're avoiding actual relative paths for easy overriding, but are using `enketo/` as a namespace basically (no `enketo` folder present necessarily), There is also a scope feature in npm `@enketo/translator` but I think that may be not be appropriate to abuse.

MartijnR:
Yes, I think I like this. And I'm changing my mind about needing to wait for a 5.0.0 version because the build system will break immediately which makes it less risky that it is overlooked.

MartijnR:
done in 01b6a80646e0598ae33383d9da60530c2f60d4a9

PR #414, opened 2017-03-29
Trigger calcUpdate() calls for changes to option labels

### LINKED ISSUE #412: jr:choice-name changes are not immediately propagated to all 2nd-level dependencies
Version: 4.27.2

From https://github.com/medic/medic-webapp/issues/3281

With the following form, change the textfield and see the new value immediately reflected in the choice labels below.  However, the bold-faced label below that will not update until the radio button is changed:

```xml
<?xml version="1.0"?>
<h:html xmlns="http://www.w3.org/2002/xforms" xmlns:ev="http://www.w3.org/2001/xml-events" xmlns:h="http://www.w3.org/1999/xhtml" xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms" xmlns:xsd="http://www.w3.org/2001/XMLSchema">
  <h:head>
    <h:title>Test Issue 3281</h:title>
    <model>
      <instance>
        <issue-3281 delimiter="#" id="issue_3281" prefix="J1!issue_3281!" version="2017-03-24">
          <input>Default Value</input>
          <translator>clinic</translator><output/><debug/>
          <note/>
          <meta>
            <instanceID/>
          </meta>
        </issue-3281>
      </instance>
      <bind nodeset="/issue-3281/input" type="string"/>
      <bind nodeset="/issue-3281/translator" type="select1"/>
      <bind calculate="jr:choice-name( /issue-3281/translator ,' /issue-3281/translator ')" nodeset="/issue-3281/output" type="string"/>
      <bind nodeset="/issue-3281/debug" readonly="true()" type="string"/>
      <bind nodeset="/issue-3281/note" readonly="true()" type="string"/>
      <bind calculate="concat('uuid:', uuid())" nodeset="/issue-3281/meta/instanceID" readonly="true()" type="string"/>
    </model>
  </h:head>
  <h:body>
    <input ref="/issue-3281/input">
      <label>Input</label>
    </input>
    <select1 ref="/issue-3281/translator">
      <label>Translator</label>
      <item>
        <label>[<output value=" /issue-3281/input "/>] District</label>
        <value>district_hospital</value>
      </item>
      <item>
        <label>[<output value=" /issue-3281/input "/>] Health Center</label>
        <value>health_center</value>
      </item>
      <item>
        <label>[<output value=" /issue-3281/input "/>] Area</label>
        <value>clinic</value>
      </item>
    </select1>
    <input ref="/issue-3281/debug">
      <label>jr:choice-name($ {translator},'$ {translator}') = **<output value=" /issue-3281/output "/>**</label>
    </input>
    <input ref="/issue-3281/note">
      <label>The value above is stale if _Input_ is changed and you leave the textbox since it modifies the choice label for _Translator_ but not the value. The choice-name function gets updated if the choice value is modified, but should also update if the choice label changes.</label>
    </input>
  </h:body>
</h:html>
```

I'm planning to dig further into this, but figured it's worth sharing here in case it's a known issue or expected behaviour.

alxndrsn:
Fixes #412 

This is a fix for the example form shown in #412

I'm not convinced this is safe or optimal for a few reasons:
* there may be other types of choice which are supported by `jr:choice-name`, but don't use `.option-wrapper`
* I'm not 100% certain how the string passed to `calcUpdate()` should be calculated, but cutting the form name off the front of the fully-qualified xpath to the model element seems to work (`owName.substring( owName.indexOf( '/', 1 ) )`)
* there _may_ be cases when an `<output>` can occur within `.option-wrapper` but outside an option label, in which case `calcUpdate()` will be triggered needlessly

Perhaps if nothing else, this fix will give a clearer insight into what the bug in #410 really is!

### DIFF
--- a/src/js/Form.js
+++ b/src/js/Form.js
@@ -1418,6 +1418,13 @@ define( function( require, exports, module ) {
                 }
                 if ( $output.text() !== val ) {
                     $output.text( val );
+
+                    // For <output>s that are children of choice-groups, trigger
+                    // dependent calculations in case one is using a jr:choice-name.
+                    var owName = $output.closest( '.option-wrapper' ).find( 'input' ).attr( 'name' );
+                    if ( owName ) {
+                        that.calcUpdate( owName.substring( owName.indexOf( '/', 1 ) ) );
+                    }
                 }
             } );
         };


MartijnR:
What an issue! :) I responded in #412.

PR #416, opened 2017-03-30
Insert required enketo-transformer version number at build-time

### LINKED ISSUE #415: Move Form.getRequiredTransformerVersion to a separate module?
It's a shame that `getRequiredTransformerVersion()` is implemented in `Form.js`, as it means enketo-core's `package.json` file is included when we browserify our code, even though this function isn't used.

If this function could be moved to its own module, then it, and `package.json`, would only be included when needed.

alxndrsn:
Closes #415

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -90,18 +90,6 @@ module.exports = function( grunt ) {
                 browsers: [ 'Chrome', 'ChromeCanary', 'Firefox', /*'Opera',*/ 'Safari' ]
             }
         },
-        replace: {
-            transformer_version: {
-                src: [ 'build/js/enketo-bundle.js' ],
-                overwrite: true,
-                replacements: [ {
-                    from: '__ENKETO_TRANSFORMER_VERSION__',
-                    to: function() {
-                        return require( './package' ).devDependencies[ 'enketo-transformer' ];
-                    }
-                } ]
-            }
-        },
         sass: {
             options: {
                 sourceMap: false,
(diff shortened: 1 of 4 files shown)


MartijnR:
This is cool stuff! Thanks! I learned something new.

My only concern is that it requires updating all build systems of apps that use Enketo Core. The regular `require('enketo-core')` will now have a broken `getRequiredTransformerVersion()` function.

alxndrsn:
Yes, it's true!  This `replace` step would also need to be implemented in downstream projects.

Some alternatives I can think of:

* keep the version number updated manually in `Form.js`
* write a similar `rewrite` Grunt task for inserting the version number into `Form.js`, and run this at appropriate times (e.g. by adding an [`npm prepublishOnly` step to `package.json`'s `scripts` section](https://docs.npmjs.com/misc/scripts)?)

MartijnR:
Let's do your:

> keep the version number updated manually in Form.js

and add a test that would immediately show in Travis when we forget to update.

alxndrsn:
Neat - I hadn't thought of writing a test for it(!)  Great idea :¬)

alxndrsn:
Updated to hard-code the expected value in `Form.js`, and test for this in `form.spec.js`.

This PR will need to be squashed when merged.

Example test failure:
```
PhantomJS 2.1.1 (Mac OS X 0.0.0) required enketo-transformer version can be obtained FAILED
	Expected '1.15.1' to be '1.15.2', 'It looks like enketo-transformer has been updated in package.json from 1.15.1 to 1.15.2.  You also need to update the value returned by From.getRequiredTransformerVersion() to the new version number.'.
```

PR #447, opened 2017-08-17
Don't do file-specific work for other types of input

alxndrsn:
There might be a subtle reason (e.g. browser compatibility) why the code was written like it was, but it seems like you shouldn't need to handle the `data-previous-file-name` `data-loaded-file-name` attributes on input types which are not `type="file"`

### DIFF
--- a/src/js/plugins.js
+++ b/src/js/plugins.js
@@ -27,6 +27,9 @@ $.fn.clearInputs = function( ev1, ev2 ) {
                 type = 'textarea';
             }
             switch ( type ) {
+                case 'file':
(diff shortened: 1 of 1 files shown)


MartijnR:
Ah yes. Not deliberate I think. Thanks. 

Let me check the effect on the new drawing/signature/annotate widget though, as some fancy wizardry with types is done there.

MartijnR:
There is indeed an issue with drawing/signature inputs, but checking this seems to be revealing more issues with clearing file inputs, so I'll fix that after merging this. Thanks!

PR #455, opened 2017-09-08
Make jr:choice-name() regex more foolproof

alxndrsn:
I'm not sure how important this is, but the old regex would not match a string like:
```
var expr = 'jr:choice-name( /something[@some_attr="val with space"] , \'/something\' )';
console.log('old', expr.match( /jr:choice-name\(([^,]+),\s?'(.*?)'\)/ ));
console.log('new', expr.match( /jr:choice-name\(\s*([^,]*[^,\s])\s*,\s*'([^']*)'\s*\)/ ));
```
I can't see any specific tests for this function, but could cook some up if that would be helpful.

### DIFF
--- a/src/js/Form.js
+++ b/src/js/Form.js
@@ -285,11 +285,11 @@ Form.prototype.replaceChoiceNameFn = function( expr, resTypeStr, selector, index
     var name;
     var $input;
     var label = '';
-    var matches = expr.match( /jr:choice-name\(([^,]+),\s?'(.*?)'\)/ );
+    var matches = expr.match( /jr:choice-name\(\s*([^,]*[^,\s])\s*,\s*'([^']*)'\s*\)/ );
 
     if ( matches ) {
         value = this.model.evaluate( matches[ 1 ], resTypeStr, selector, index, tryNative );
-        name = matches[ 2 ].trim();
+        name = matches[ 2 ];
         $input = this.view.$.find( '[name="' + name + '"]' );
 
         if ( $input.length > 0 && $input.prop( 'nodeName' ).toLowerCase() === 'select' ) {


src/js/Form.js
```diff
@@ -285,11 +285,11 @@ Form.prototype.replaceChoiceNameFn = function( expr, resTypeStr, selector, index
     var name;
     var $input;
     var label = '';
-    var matches = expr.match( /jr:choice-name\(([^,]+),\s?'(.*?)'\)/ );
```
alxndrsn: What did the question mark in `(.*?)` do?!

alxndrsn:
Also I think both implementations will ignore `jr:choice-name()` if the string argument uses `"` instead of `'`...

MartijnR:
Thanks! My `(.*?)` doesn't make sense to me either...

Your regex seems like a big improvement. Not sure what `[^,\s]` does.

It would be great to add a test at least for `'jr:choice-name( /something[@some_attr="val with space"] , \'/something\' )'`. Do you have any interest in adding one? I think it will require an XForm to be added with 1 question and some manual model edits.

MartijnR:
Hi alxndrsn, I saw those test commits pass by (yay!), but they're not in the PR yet. Not sure if the review was requested recently or if it's the old request. Just checking if they are in progress or just need to be added to your `safer-choice-name` branch.

MartijnR:
Oh never mind, I see it's wip in https://github.com/enketo/enketo-core/pull/475. Will delete my comments now.

alxndrsn:
Looking at this again, `jr:choice-name( /something[@some_attr="val with space"] , \'/something\' )` seems _really_ contrived.  I think the other PR has a reasonable amount of flexibility added to the regex without getting silly.

PR #467, opened 2017-11-04
Remove unused argument

alxndrsn:
Doesn't look like the second arg is used.  If it's being used by dependent projects, we should document that in `file-manager`.


MartijnR:
Good point.  It's used to pass an override filename (for iOS).

So this should be documented in file-manager or that filename rename functionality should be added to the file-manager. I'll create an issue.

PR #470, opened 2017-11-04
Expose fileManager.isTooLarge()

### LINKED ISSUE #465: Limit file upload size on client-side
I know there are good reasons for validating file upload size on the server-side as well as the client-side, but for usability I'd like to check the size of a file upload before saving a form.

I have a form definition for collecting images:
```
<h:html xmlns="http://www.w3.org/2002/xforms"                                                                                                                                                xmlns:h="http://www.w3.org/1999/xhtml"                                                                                                                                               xmlns:ev="http://www.w3.org/2001/xml-events"
        xmlns:xsd="http://www.w3.org/2001/XMLSchema"
        xmlns:jr="http://openrosa.org/javarosa">
    <h:head>
        <h:title>Photos - Demo Form</h:title>

        <model>
            <instance>
                <photos id="photos">
                    <my_photo/>
                    <meta>
                        <instanceID/>
                    </meta>
                </photos>
            </instance>

            <bind nodeset="/photos/my_photo" type="binary"/>
        </model>
    </h:head>

    <h:body>
        <upload ref="/photos/my_photo" mediatype="image/*">
            <hint>Select a picture or take a photo</hint>
            <label>Image widget</label>
        </upload>
    </h:body>
</h:html>
```

# Things I've tried

## 1. Override `FormModel.prototype.types.binary.validate`

Unfortunately due to https://github.com/enketo/enketo-core/blob/master/src/js/Form-model.js#L1565-L1569, validate() is never actually called for `binary` input fields.

Even if it _were_ called, `validate()` is only passed the "value" of the field, which for an `<input type="file">` is actually the _name_ of the file.  What we would actually need access to is `that.nodes[0].files` to access e.g. `that.nodes[0].files[0].size`

## 2. Add a photo input widget

Adding custom validation in a widget:

```
    …
    Photowidget.prototype._init = function() {
        var $el = $( this.element );
        $el.on('change', function() {
            var i, tooBig, $this;

            for ( i = 0; i < this.files.length; ++i ) {
                if ( this.files[ i ].size > MAX_FILE_SIZE ) {
                    tooBig = true;
                }
            }

            if ( tooBig ) {
                …TODO mark this field as invalid so that constraint message is displayed…
            }
        });
    };
    …
    module.exports = {
        'name': pluginName,
        'selector': 'input[type=file][accept="image/*"]',
    };
    …
```
Adding a warning in HTML from the widget seems trivial, but there doesn't seem to be a simple way of invalidating the corresponding field in the model.

-----

1. Is there a simple way to impose a size constraint on a binary field in the current version of Enketo?
2. If not, do you have any suggestions for how it might be achievable?

alxndrsn:
Closes #465

### DIFF
--- a/src/js/file-manager.js
+++ b/src/js/file-manager.js
@@ -12,36 +12,39 @@ var Promise = require( 'lie' );
 var $ = require( 'jquery' );
 var utils = require( './utils' );
 
-var supported = typeof FileReader !== 'undefined',
-    notSupportedAdvisoryMsg = '';
+var fileManager = {};
+
+var supported = typeof FileReader !== 'undefined';
+
+fileManager.notSupportedAdvisoryMsg = '';
 
 /**
  * Initialize the file manager .
  * @return {[type]} promise boolean or rejection with Error
  */
-function init() {
-    if ( supported ) {
+fileManager.init = function() {
+    if ( fileManager.isSupported() ) {
         return Promise.resolve( true );
     } else {
         return Promise.reject( new Error( 'FileReader not supported.' ) );
     }
-}
+};
 
 /**
  * Whether filemanager is supported in browser
  * @return {Boolean}
  */
-function isSupported() {
+fileManager.isSupported = function() {
     return supported;
-}
+};
 
 /**
  * Whether the filemanager is waiting for user permissions
  * @return {Boolean} [description]
  */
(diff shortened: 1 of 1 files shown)


MartijnR:
Great! Thanks!

PR #472, opened 2017-11-04
Test on travis on more node versions

alxndrsn:
Is there any value to testing against a specific version?  If not, then maybe we should switch from `6` to long term support (`lts/*`) or latest stable (`node`).  Or perhaps supporting a very old version will make browser compatibility a bit safer.  Any thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 dist: trusty
 language: node_js
 node_js:
-  - "6"
+  - lts/*
 addons:
   apt:
     packages:


MartijnR:
Having one value might be sufficient because nodeJS just needs to run the test server and compile files. The tests themselves run in (latest stable) headless Chrome now. It's not important to test on multiple nodeJS versions imo. So just `lts/*` sounds good to me.

alxndrsn:
Updated to just test on node LTS version.

PR #493, opened 2017-12-22
Handle complex jr:choice-name() better

### LINKED ISSUE #490: recent jr:choicename-regression
https://github.com/kobotoolbox/enketo-express/issues/900

alxndrsn:
Fix regression in jr:choice-name() when the function call is embedded within
other function calls.

Closes #490 

@MartijnR I'm not really happy with either of the new tests supplied here:

1. the mocking required for the "unit" test is horrible
2. the integration test is probably better, but it seems a bit much that we have to include a new form just to test changes to a regex

One option to clean things up would be to expose the regex somehow, e.g. include it in a different file and `require(…)` it into both `Form.js` and a test file.  Seems a bit extreme to have a separate file for a single regex, though.

But if you're happy with the PR as-is, please merge.

### DIFF
--- a/src/js/Form.js
+++ b/src/js/Form.js
@@ -292,8 +292,8 @@ Form.prototype.replaceChoiceNameFn = function( expr, resTypeStr, selector, index
     var $input;
     var label = '';
     var matches = expr.match( new RegExp( 'jr:choice-name\\(([^,]+),\\s*(?:' +
-        '"(.*)"|' +
-        "'(.*)'" +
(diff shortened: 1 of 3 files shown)


MartijnR:
Thanks so much for this PR!! 

Yes, we could make it easier to test the regex by exposing it (declaring the regex var at the top of Form.js and then making it available as a static method like `Form.requiredTransformerVersion`, so you don't have to instantiate a Form instance to obtain it in the tests). I will merge this first for now.

PR #524, opened 2018-07-04
Fix spelling mistake in comment

alxndrsn:


MartijnR:
Ah, must be the Dutch trying to break through! 👍

PR #525, opened 2018-07-04
Add missing quote mark in example code in comment

alxndrsn:
...or maybe the other quote mark should be removed? 🙂


MartijnR:
👍 :)

PR #628, opened 2019-05-22
fixed: in Safari and all iOS browsers, scrolling to top on scrollable…

### LINKED ISSUE #626: `pageEl.scrollIntoView` does not work properly on iOS
In a previous issue (#331), it looks like you added `pageEl.scrollIntoView` to ensure that the form scrolls to the top on page flip. However, it looks like this is broken on iOS. In particular:
- in a regular browser (e.g. Safari), the header is hidden. See attached video comparing the way the same form is loaded on iOS and android
[scrollIntoViewAndroidvsiOS.mp4.gz](https://github.com/enketo/enketo-core/files/3180427/scrollIntoViewAndroidvsiOS.mp4.gz)
- that may be bearable, but if embedded into a more complex single screen app, the additional scroll breaks the entire page because it offsets the body (see https://github.com/e-mission/e-mission-phone/pull/563#issuecomment-492359427) and previous.

The commit for #331 added `scrollIntoView` in three locations - commenting just one of them out - the one with `pageEl` seems to fix everything.
https://github.com/e-mission/e-mission-phone/pull/563#issuecomment-492403123

Since I am manually editing the generated bundle anyway (https://github.com/enketo/enketo-core/issues/623), I commented out that line, so I am not blocked on this. But I thought it might be helpful to file an issue in case you want a longer-term fix.

I am not sure what would be a great fix though.

MartijnR:
… new page hides form header, closes #626

### DIFF
--- a/src/js/page.js
+++ b/src/js/page.js
@@ -299,7 +299,8 @@ export default {
             .eq( 0 )
             .trigger( 'fakefocus' );
 
-        pageEl.scrollIntoView();
+        const scrollToElement = this.form.view.html.parentElement.querySelector( '.form-header' ) || pageEl;
+        scrollToElement.scrollIntoView();
     },
     _toggleButtons( index ) {
         const i = index || this._getCurrentIndex(),
--- a/src/js/page.js
+++ b/src/js/page.js
@@ -299,8 +299,8 @@ export default {
             .eq( 0 )
             .trigger( 'fakefocus' );
 
-        const body = this.form.view.html.closest( 'body' );
-        const header = body ? body.querySelector( '.form-header' ) : null;
+        const wrapper = this.form.view.html.parentElement;
+        const header = wrapper ? wrapper.querySelector( '.form-header' ) : null;
         ( header || pageEl ).scrollIntoView();
     },
     _toggleButtons( index ) {
--- a/src/js/page.js
+++ b/src/js/page.js
@@ -299,8 +299,9 @@ export default {
             .eq( 0 )
             .trigger( 'fakefocus' );
 
-        const scrollToElement = this.form.view.html.parentElement.querySelector( '.form-header' ) || pageEl;
-        scrollToElement.scrollIntoView();
+        const body = this.form.view.html.closest( 'body' );
+        const header = body ? body.querySelector( '.form-header' ) : null;
+        ( header || pageEl ).scrollIntoView();
     },
     _toggleButtons( index ) {
         const i = index || this._getCurrentIndex(),


MartijnR:
Hi @shankari, 

Thanks for all your help! Does this PR solve issue #626 in a satisfactory manner?

I think it makes sense to normally scroll to a form header ( I added a fallback to the current behavior if a `.form-header` is missing or cannot be found).

This does make me want to rethink whether it is still desirable to show the form title on every page, but that will be a different issue.

shankari:
Testing to see if this fixes my issue.

shankari:
so one high level issue that I see is with embedded surveys. If I have a single page app (e.g. angular) with multiple views and I have embedded different surveys into different tabs (e.g. one survey for profile and one survey for incident), then this will break because there will be two `form-header` elements in the body. Note that angular ui router (at least v1) keeps inactive tabs around for faster loading.

Table showing a single page app with multiple tabs. Note that the tab views are present in the DOM even though they are not displayed.

| Displayed tab (diary) | Hidden tab (Big data) | Hidden tab (Recommendations) |
|--------------------- | --------------------- | --------------------------------- |
| Screen Shot 2019-05-22 at 3 04 24 PM | Screen Shot 2019-05-22 at 3 04 38 PM | Screen Shot 2019-05-22 at 3 04 54 PM

MartijnR:
Thanks for reviewing. How about a more restricted version where we replace:

```js
        const body = this.form.view.html.closest( 'body' );
        const header = body ? body.querySelector( '.form-header' ) : null;
```

with 

```js
        const wrapper = this.form.view.html.parentElement;
        const header = wrapper ? wrapper.querySelector( '.form-header' ) : null;
```

shankari:
That should work. Let me test.

MartijnR:
Thanks. I've pushed it quickly.

shankari:
This does not fix the issue with the embedded app 
https://github.com/e-mission/e-mission-phone/pull/563#issuecomment-492359427
see attached video.
[display_moves_up_again.mp4.gz](https://github.com/enketo/enketo-core/files/3209625/display_moves_up_again.mp4.gz)

It seems to be reasonable in Safari on OSX although I can still see a slight displacement from the stop of the page - the gray border at the top goes missing. See screenshot below.

| Hidden top border | After scrolling up |
|---------------------|-----------------|
| Screen Shot 2019-05-22 at 4 09 36 PM | Screen Shot 2019-05-22 at 4 10 10 PM

shankari:
The real issue for me is the embedded app and the resulting displacement.

MartijnR:
Good point about the border. Thanks!

The video doesn't really show me the issue with the embedded app.

MartijnR:
Closing this, as I do not see how we can devote resource to this. I shouldn't have tried. Sorry about that.

PR #629, opened 2019-05-26
Duplicate word on string 175

kant:


MartijnR:
Thanks for taking the trouble to send this PR! I sat on it, as we were in the middle of moving documentation around. I don't normally do this but I made the change myself now in the above commit instead of asking you re-do the single-word-deletion.

PR #632, opened 2019-06-03
changed: reorganized tutorials

MartijnR:
Reorganized tutorials a bit. It looks like we cannot easily get rid of the 01- path prefixes, which is unfortunate.


MartijnR:
Also added missing pages. Goal was to just get them there and worry about the correct jsdoc syntax later, though please let me know if you see any changes that should have been documented differently!

PR #633, opened 2019-06-12
fixed: Grid theme options with appearance horizontal break out of cell

### LINKED ISSUE #546: Grid theme options with appearance columns/columns-pack break out of cell
_From @MartijnR on July 22, 2015 22:14_

screen shot 2015-07-22 at 4 12 38 pm

css `break-word: breakall` solves this but may need to reduce margins and padding to give a little more space.

Not related to ltr/rtl.

_Copied from original issue: kobotoolbox/enketo-express#265_

gushil:
Trying to fix #546

### DIFF
--- a/src/sass/core/_main.scss
+++ b/src/sass/core/_main.scss
@@ -337,7 +337,7 @@ legend {
         display: block;
         margin: 0;
         cursor: pointer;
-        padding: 4px 0;
+        padding: 4px;
         margin: 0 8px 1px 10px; // TODO: remove hover rule for tables (in likert already done)
 
         &:hover:not(.filler) {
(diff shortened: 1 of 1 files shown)


gushil:
> Thanks gushil. The change from `margin: 4px` to `margin: 4px 0`, doesn't look so good on hover. Best to revert to original.
> 
> Here's a screenshot for the regular theme (but same issue in Grid theme)
> 
> Screen Shot 2019-06-24 at 1 13 13 PM
> 
> I think if you revert that all should be good!

I think the changes you meant is `padding: 4px 0`, am I right? 
That one is reverted in new commit. 
Thanks.

Issue #578, opened 2018-10-26
Add support for max-pixels attribute


MartijnR:
also requires some work in enketo-xslt

Feature: https://opendatakit.github.io/xforms-spec/#bind-attributes

MartijnR:
Detailed guidance to serve as exercise:

-1. Get enketo-core running locally
0. [This test form](https://docs.google.com/spreadsheets/d/1SA2mxwGKPSgsxzeevyGHdc-Kb_fWDwWQewfcLNehhSc/edit?usp=sharing) can be used. Download as XLSX and convert to XForm [here](https://opendatakit.org/xlsform/). Place XForm (image-upload.xml) in your local `/test/temp` folder. Run the form locally with http://localhost:8005/?xform=image-upload.xml.
1. All work (probably) to be done in `/src/js/file` folder with a few lines in file-manager.js
2. Use hardcoded value of 1024 pixels (will be trivial to change that value to form-defined value later, not part of this exercise).
3. Upon completion, when uploading a high resolution image in the form and clicking validate the file will have been transformed according to the spec.
4. Feel free to just use any hack to get fileManager.getCurrentFiles() to return your transformed file. That's not important at this stage.
5. Ask questions whenever anything is unclear!

MartijnR:
From enketo-transformer 1.32.0 onwards, the following output is created if max-pixels was defined:

`<input type="file" data-max-pixels="100" />`

test form: https://docs.google.com/spreadsheets/d/1ZdOqWxXstfR4y-dzWgThsSduJ0imsnon9vFbKwNu_3E/edit?usp=sharing

PR #636, opened 2019-06-13
added: Support for max-pixels attribute

gushil:
#578


src/js/file-manager.js
```diff
@@ -163,4 +171,68 @@ fileManager.isTooLarge = () => { return false; };
...
 
+/**
+ * @function getMaxImageWidth
+ *
+ * @description Replace with function that determines maximum image width.
+ *
+ * @return {number} maximum image width
+ */
+fileManager.getMaxImageWidth = () => { return 1024; };
+
```
MartijnR: get maxPixels from this.props in widget. You can set this to 1024 for now (we'll need a small change in enketo-xslt to get the form-defined value).

src/js/file-manager.js
```diff
@@ -163,4 +171,68 @@ fileManager.isTooLarge = () => { return false; };
...
+/**
+ * @function isImageFile
+ *
+ * @description Check whether current file is image by reading its type.
+ *
+ * @return {boolean} whether file is image
+ */
+fileManager.isImageFile = file => { return file && file.type.split( '/' )[ 0 ] === 'image'; };
+
```
MartijnR: This method won't be necessary after the above. The widget is aware of the type.

src/js/file-manager.js
```diff
@@ -163,4 +171,68 @@ fileManager.isTooLarge = () => { return false; };
...
+ * @function resizeImage
+ *
+ * @param {File} file - image file to be resized
+ * @param {number} maxWidth - maximum width of resized image
+ * @param {number} maxHeight - maximum height of resized image
+ *
+ * @return {Promise<Blob>} promise of resized image blob
+ */
+fileManager.resizeImage = ( file, maxWidth, maxHeight ) => {
+    return new Promise( ( resolve, reject ) => {
```
MartijnR: I think the method should be called from the widget.

src/js/file-manager.js
```diff
@@ -134,11 +134,19 @@ fileManager.getCurrentFiles = () => {
...
+                    files.push( file );
+                } );
+            } else {
+                file = new Blob( [ file ], {
+                    type: file.type
+                } );
+                file.name = newFilename;
+                files.push( file );
+            }
         }
```
MartijnR: I think none of this is required. See below.

src/js/file-manager.js
gushil: Done. Now got maxPixels from `filepicker.js` `get props()`

src/js/file-manager.js
gushil: Done. Delete unnecessary methods in `fileManager`

src/js/file-manager.js
gushil: Method now called from `Filepicker`.
Little question. What should I do when resizeImage is successfull?

src/js/file-manager.js
```diff
@@ -163,4 +163,50 @@ fileManager.isTooLarge = () => { return false; };
...
+            let height = image.height;
+
+            if ( width <= maxWidth && height <= maxHeight ) {
+                resolve( file );
+            }
+
+            let newWidth;
+            let newHeight;
+
+            if ( width > height ) {
```
MartijnR: If I understand correctly, this means that the two parameters of this function cannot actually be used. That is fine, as we don't need 2. So I think it would simplify a bit by changing this function to only have 1 `maxPixel` parameter.

src/js/file-manager.js
```diff
@@ -163,4 +163,50 @@ fileManager.isTooLarge = () => { return false; };
  */
 fileManager.getMaxSizeReadable = () => { return `${5}MB`; };
 
+/**
+ * @function resizeImage
```
MartijnR: We actually create a replacement for `file-manager.js` in Enketo Express. So let's move this method elsewhere. I'm thinking in `utils.js`.

src/widget/file/filepicker.js
```diff
@@ -241,6 +242,15 @@ class Filepicker extends Widget {
         }
     }
 
+    _resizeImage( file, mediaType ) {
+        if ( mediaType !== 'image/*' ) {
```
MartijnR: maybe also add a check to see if this.props.maxPixels is defined (to be ready for future)

src/js/file-manager.js
MartijnR: Good question. I haven't figured this out myself. Unfortunately, it is not possible to dynamically replace the file in a `<input type="file">`.


I think we have make a tweak to `getCurrentFiles` in the file-manager, and in the filepicker add an attribute to the file input if it was resized. So if that attribute exists getCurrentFiles will not use `this.files[0]` to get the file but gets the resized blob instead (not sure how though...). We do something a little similar with the drawing widget/canvas. 

It would also be good to use the resized blob for the preview, so users don't get a false sense of picture quality to be uploaded, if the quality was actually visibly reduced a lot.

MartijnR:
@gushil, would you like me to make the little tweak to enketo-xslt (and enketo-transformer) to pass the max-pixel value set in the XForm, or are you interested in doing this?

I'm asking because xsl is such an obscure language.

gushil:
> @gushil, would you like me to make the little tweak to enketo-xslt (and enketo-transformer) to pass the max-pixel value set in the XForm, or are you interested in doing this?
> 
> I'm asking because xsl is such an obscure language.

I think for now it would be best if you do the tweak so I can focus on this issue. Is it ok?

gushil:
`resizeImage` has been moved into `utils.js`.
Now still working on how to save resized image blob. Thinking maybe saving into localstorage?

MartijnR:
I will make the enketo-xslt change and let you know when done.

> Thinking maybe saving into localstorage? 

Would like to avoid a browser db solution if we can. Could we use the preview perhaps? (objectURl or dataURI?)

MartijnR:
done, see https://github.com/enketo/enketo-core/issues/578#issuecomment-507350216

src/js/file-manager.js
gushil: Now image preview and getCurrentFiles using resized image file.

src/js/file-manager.js
gushil: Done

src/js/file-manager.js
gushil: `resizeImage` is moved into `utils.js`

src/widget/file/filepicker.js
gushil: Checking added.

src/widget/file/filepicker.js
```diff
@@ -241,6 +255,25 @@ class Filepicker extends Widget {
...
     }
 
+    _resizeFile( file, mediaType ) {
+        return new Promise( ( resolve, reject ) => {
+            if ( mediaType !== 'image/*' ) {
+                reject( file );
+            }
+
+            // file is image, resize it
+            if ( this.props && this.props.maxPixels ) {
```
MartijnR: This function will now not resolve or reject if the condition is not truthy, right (e.g. if maxPixels was not defined in the XForm)? It will remain pending, and `finally()` will never be reached.

src/widget/file/filepicker.js
```diff
@@ -257,6 +290,7 @@ class Filepicker extends Widget {
     get props() {
         const props = this._props;
         props.mediaType = this.element.getAttribute( 'accept' );
+        props.maxPixels = this.element.dataset.maxPixels ? this.element.dataset.maxPixels : 1024;
```
MartijnR: no default value. The default is that no resizing will happen.

PR #637, opened 2019-06-17
fixed: iOS 12 browsers mess up drawing widget when rotating device

### LINKED ISSUE #625: iOS 12 browsers mess up drawing widget when rotating device
Try switching orientation and adding to the drawing in both orientations.

gushil:
fixed #625

### DIFF
--- a/src/widget/draw/draw-widget.js
+++ b/src/widget/draw/draw-widget.js
@@ -394,6 +394,11 @@ class DrawWidget extends Widget {
             that._forceUpdate();
             that._resizeCanvas( canvas );
         } );
+
+        $( window ).on( 'orientationchange', () => {
+            that.disable();
+            setTimeout( that.enable.bind( that ), 500 );
+        } );
     }
 
     // Adjust canvas coordinate space taking into account pixel ratio,


gushil:
> Hi @gushil,
> 
> Not sure if I forgot to post this earlier (or if I just cannot find it), but I was wondering if you tried using the same as for _handleResize() on `orientationchange`. It would be more consistent and I'd like to avoid those timeout solutions if we can.
> 
> ie.
> 
> ```js
> that._forceUpdate();
> that._resizeCanvas( canvas );
> ```

Actually, I've tried a couple of things (when `orientationchange` occurred):
1. Combination of `_forceUpdate()` and `_resizeCanvas(  canvas )` you mentioned above.
2. Made `disable()` into function that returns Promise and call `enable()` after that.
3. Calling `enable()` with some `setTimeOut` after `disable()` .

Unfortunately, with Chrome on iOS, no.3 is the one with desired result.

MartijnR:
Thanks! Will sit on this for a little bit. It looks like Chrome on iOS is using the same rendering engine as Safari (and not the old one), so I want to see what happens in the next Chrome for iOS release.

MartijnR:
Looks like this is not an issue any more or not so important

Issue #510, opened 2018-02-27
Show warning if select_multiple value has spaces


MartijnR:
Pyxform will check for this but there is always a chance that a form that builds a choice list from external data contains such options, so it needs to be checked on the client.

https://github.com/opendatakit/javarosa/issues/263#issuecomment-369019465

theywa:
@MartijnR, Can you elaborate more about the "builds a choice list from external data contains" and "checked on the client"? and if possible steps to testing it

MartijnR:
hi @theywa, 

Enketo Express can load external data files that will be used in the XForm (referred to with `<instance src="..."/>`) (e.g. clinicaldata.xml, users.xml as OC uses). One way they can be used is to build choice lists for select_multiple questions. Since form validators do not know the content of this external data file (which can change as well), they cannot show warnings for values that contain spaces (and are not allowed). 

In Enketo Core, the ability to load external files is not included. Instead the XML content of such external data files can be passed during form instantiation with the data property. 

I'll create a form and a form instantiation snippet you could use to develop this check (and create an automated test from).

Let's add this feature only to itemset.js (i.e. dynamic choice list generator for both internal and external data). So it won't be available to hardcoded non-dynamic choice lists (we can rely on form validators to check for those). I think the only feasible way to do this is to add the check at the time a new dynamic list is built, so that means that a warning may only pop up halfway through the data entry session.

MartijnR:
Here is a test form (added .txt to extension just to fool github) which you can add to the /test/forms folder:

[external-data.xml.txt](https://github.com/enketo/enketo-core/files/3378387/external-data.xml.txt)

To instantiate with external data in app.js (temporarily) add:

```js
modelStr: modelStr,
        external: [ {
            id: 'mydata',
            xml: new DOMParser().parseFromString(
                `   <root>
                        <item>
                            <label>11</label>
                            <group>a</group>
                            <name>1 1</name>                        
                        </item>
                        <item>
                            <label>22</label>
                            <group>a</group>
                            <name>22</name>
                        </item>
                        <item>
                            <label>33</label>
                            <group>b</group>
                            <name>33</name>
                        </item>
                        <item>
                            <label>44 (value has spaces!)</label>
                            <group>b</group>
                            <name>4 4</name>
                        </item>
                    </root>
                `, 'text/xml' )
        } ]
```

[...]

PR #643, opened 2019-07-22
added: Show warning if select_multiple value has spaces

theywa:
[510](https://github.com/enketo/enketo-core/issues/510) 
Preview 
Screenshot from 2019-07-22 18-02-16
I change the color to highlight the warning.

### DIFF
--- a/src/js/fake-translator.js
+++ b/src/js/fake-translator.js
@@ -74,6 +74,9 @@ const SOURCE_STRINGS = {
     'alert': {
         'gotonotfound': {
             'msg': 'Failed to find question \'__path__\' in form. Is it a valid path?'
+        },
+        'valuehasspaces': {
+            'multiple': 'Select multiple question has an illegal value "__value__" that contains a space.'
         }
     },
     'confirm': {
--- a/src/js/itemset.js
+++ b/src/js/itemset.js
@@ -6,6 +6,8 @@
 
 import $ from 'jquery';
 import { parseFunctionFromExpression } from './utils';
+import dialog from 'enketo/dialog';
+import { t } from 'enketo/translator';
 
 export default {
     /**
@@ -168,7 +170,13 @@ export default {
                 }
                 // Obtain the value of the secondary instance item found.
                 const value = that.getNodeFromItem( valueRef, item ).textContent;
-
+                /**
+                 * #510 Show warning if select_multiple value has spaces
+                 */
+                const multiple = ( inputAttributes[ 'data-type-xml' ] == 'select' ) && ( inputAttributes[ 'type' ] == 'checkbox' ) || ( $list[ 0 ] && $list[ 0 ].multiple );
+                if ( multiple && ( value.indexOf( ' ' ) > -1 ) ) {
+                    dialog.alert( t( 'alert.valuehasspaces.multiple', { value: value } ) );
+                }
                 if ( templateNodeName === 'label' ) {
                     optionsFragment.appendChild( that.createInput( inputAttributes, translations, value ) );
                 } else if ( templateNodeName === 'option' ) {
--- a/test/forms/external-data-v2.xml
+++ b/test/forms/external-data-v2.xml
@@ -1,0 +1,89 @@
+<?xml version="1.0"?>
+<h:html xmlns="http://www.w3.org/2002/xforms" 
+    xmlns:ev="http://www.w3.org/2001/xml-events" 
+    xmlns:h="http://www.w3.org/1999/xhtml" 
+    xmlns:jr="http://openrosa.org/javarosa" 
+    xmlns:odk="http://www.opendatakit.org/xforms" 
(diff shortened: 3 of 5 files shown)


theywa:
@MartijnR - I update the PR also add the test forms but not finish the automated test(need to learn it first and I think it'll take some time so I skip it first and will back to try it again later). Where should I put the automated code in itemset.spec.js or form.spec.js? Also when I tried running the test(npm test) locally, I see some error related to the Google Maps and some other validation error that I think need to fix first before trying to add new test.
Preview:
Screenshot from 2019-07-23 14-39-41
Screenshot from 2019-07-23 14-39-03

src/js/fake-translator.js
```diff
@@ -57,7 +57,11 @@ const SOURCE_STRINGS = {
     },
     'selectpicker': {
         'noneselected': 'none selected',
-        'numberselected': '__number__ selected'
+        'numberselected': '__number__ selected',
+        'valuehasspaces': {
+            'minimal': 'select one question has an illegal value __value__ that contains a space',
```
MartijnR: Since these messages can also be shown without a widget (a regular select multiple e.g.), let's put the 'valuehasspaces' key under 'alert' instead (so 'alert.valuehasspaces.multiple').

src/js/itemset.js
```diff
@@ -244,6 +246,12 @@ export default {
         const option = document.createElement( 'option' );
         option.textContent = label;
         option.value = value;
+        /**
+         * #510 Show warning if select_one value has spaces
```
MartijnR: Let's actually not do this check for select_one. Sorry for the confusion. I was't quite sure either (technically it's not wrong to have a space in a select_one value), and just saw that ODK Collect seems to also allow this.

MartijnR: The test form (v2) is still good as it contains 3 select_multiple questions that prompt erros.

MartijnR: P.S. `createOption` is not only for select_one. It is used for both multiple and single selects.

src/js/itemset.js
```diff
@@ -274,6 +282,12 @@ export default {
         } );
         input.value = value;
         label.appendChild( input );
+        /**
+         * #510 Show warning if select_multiple value has spaces
+         */
+        if ( ( attributes[ 'data-type-xml' ] == 'select' ) && ( attributes[ 'type' ] == 'checkbox' ) && ( value.indexOf( ' ' ) > -1 ) ) {
+            dialog.alert( t( 'selectpicker.valuehasspaces.multiple', { value: value } ) );
+        }
```
MartijnR: I think it may be better to move your check to before `createOption` or `createInput` are called. Do you agree?

test/forms/external-data.xml
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0"?>
```
MartijnR: This file can be removed I think, since it is superseded by v2 above.

src/js/fake-translator.js
MartijnR: Would be nice to place __value__ in between double quotation marks and to start sentence with Capital and end with a point.

theywa: added double quotation marks

src/js/itemset.js
theywa: code removed

src/js/itemset.js
theywa: sure, update code position

test/forms/external-data.xml
theywa: file removed

src/js/itemset.js
MartijnR: The remaining issue is that it doesn't for select_multiple with a minimal appearance. I tested it by removing the other 2 `<select>` (without minimal appearance) from the XForm.

MartijnR:
Thanks @theywa. Will wait for tests and final fix, before merging!

src/js/itemset.js
theywa: update the condition to include select_multiple with a minimal appearance.

theywa:
> Thanks @theywa. Will wait for tests and final fix, before merging!

I just added the test

src/js/itemset.js
MartijnR: Thanks @theywa. It works but now it also catches a single select with minimal (try by changing `<select>` to `<select1>` in the xform and it also catches autocomplete select_one questions.

I think this would work well instead:

```js
const multiple = ( inputAttributes[ 'data-type-xml' ] == 'select' && inputAttributes[ 'type' ] == 'checkbox' ) || ( $list[ 0 ] && $list[ 0 ].multiple );
```

theywa: @MartijnR - Yes, you're right. I miss checking minimal for select1, will be careful next time. I just update the code and how about tests? is it okay like that?

PR #645, opened 2019-08-07
updated: Show warning if select_multiple value has spaces

theywa:
feedback testing on enketo-express: join warning messages into one


MartijnR:
Thanks! 

Jsbeautifier has an issue, but will merge this time, and fix that issue in another commit.

PR #650, opened 2019-08-16
Remove old docs before building new ones

magicznyleszek:
I discovered that `jsdoc` generator is not aggressive and leaves leftovers. Now we delete old `docs` before generating new ones (this change will be included in all other repos PRs for documentation)


MartijnR:
Thanks!

PR #652, opened 2019-08-17
fixed: fileManager.getCurrentFiles using asynchronous code for synchronous result( offline mode )

theywa:
[#578](https://github.com/enketo/enketo-core/issues/578)


MartijnR:
👍

PR #658, opened 2019-08-30
Further update docs

magicznyleszek:
Sorry this is so big ⛰ It is still not 100% documented, as some functions could use descriptions etc. I managed to document `types.js`, but I'm not super happy with the output - but this is how internet is suggesting (I tried some other ideas, but only this one worked). Besides adding bunch of JSDoc comments, I also excluded two hardcoded libraries from coverage, thus increasing it by ~3% 🎆


src/js/form.js
```diff
@@ -251,6 +301,10 @@ Form.prototype.init = function() {
     return loadErrors;
 };
 
+/**
+ * @param {string} xpath
+ * @return {Array<Error>} A list of errors originated from `goToTarget`. Empty if everything went fine.
+ */
```
MartijnR: This is an array of error messages (strings) not Error instances

src/js/form-model.js
```diff
@@ -86,6 +82,8 @@ FormModel.prototype = {
 
 /**
  * Initializes FormModel
+ *
+ * @return {Array<Error>} list of initialization errors
```
MartijnR: This is an array of error messages (strings) not Error instances

src/js/form.js
```diff
@@ -383,9 +440,12 @@ Form.prototype.getModelValue = function( $control ) {
...
  *
- * @param  {string} attr - The attribute name to search for
- * @param  {?string} filter - The optional filter to append to each selector
- * @param  {{nodes:Array<string>=, repeatPath: string=, repeatIndex: number=}=} updated - The object containing info on updated data nodes
+ * @param {string} attr - The attribute name to search for
+ * @param {string} [filter] - The optional filter to append to each selector
+ * @param {object} [updated] - The object containing info on updated data nodes
+ * @param {Array<string>} [updated.nodes]
+ * @param {string} [updated.repeatPath]
+ * @param {number} [updated.repeatIndex]
```
MartijnR: I think it might be good to define the updated object centrally, since this is such a key part of the engine, and occurs many times (calculate.js, itemset.js, relevant.js, repeat.js, required.js, output.js, readonly.js). What do you think?

MartijnR:
In the future, it would be better to leave out the generated docs for easier review (but I know this can still be considered the 'initial' docs version).

src/js/form.js
```diff
@@ -135,6 +184,7 @@ Form.prototype.addModule = function( module ) {
  *
  * Initializes the Form instance (XML Model and HTML View).
  *
+ * @return {Array<Error>} List of initialization errors.
```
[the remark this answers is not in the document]
magicznyleszek: Fixed

src/js/form.js
magicznyleszek: Fixed

src/js/form-model.js
magicznyleszek: Fixed

src/js/form.js
magicznyleszek: You're right. I created the `@typedef` for `UpdatedDataNodes`

Issue #527, opened 2018-07-09
range widget: rating appearance


MartijnR:
https://github.com/XLSForm/pyxform/issues/145

https://github.com/XLSForm/xlsform.github.io/issues/128

theywa:
@MartijnR - is the new appearance(rating) will look like the one in the [doc](https://docs.opendatakit.org/form-question-types/#range-widget-with-rating)?

MartijnR:
Yes!

MartijnR:
test form: https://docs.google.com/spreadsheets/d/1_2ka0cRBi8M-kVyfEkNe2OTN2PwzH0tY1DO_t0iL_YE/edit?usp=sharing

MartijnR:
Martijn's to check list:

* [ ] step of 2, 3
* [ ] readonly
* [ ] programmatically emptying (calculate outcome is '')
* [ ] clearing by user

PR #659, opened 2019-09-04
add: range widget(rating appearance)

theywa:
#527 
it looks like this
Screenshot from 2019-09-04 17-24-07

### DIFF
--- a/src/widget/range/range-widget.js
+++ b/src/widget/range/range-widget.js
@@ -6,9 +6,11 @@ import events from '../../js/event';
  * @extends Widget
  */
 class RangeWidget extends Widget {
-
+    /**
+     * @type string
+     */
     static get selector() {
-        return '.or-appearance-distress input[type="number"], .question:not(.or-appearance-analog-scale) > input[type="number"][min][max][step]';
+        return '.or-appearance-distress input[type="number"], .question:not(.or-appearance-analog-scale):not(.or-appearance-rating) > input[type="number"][min][max][step]';
     }
 
     _init() {
@@ -75,6 +77,8 @@ class RangeWidget extends Widget {
 
     /**
      * This is separated so it can be extended (in the analog-scale widget)
+     *
+     * @return {string} HTML string
      */
     _getHtmlStr() {
         const html =
@@ -99,12 +103,19 @@ class RangeWidget extends Widget {
         return html;
     }
 
+    /**
+     * @param {number} completeness
+     */
     _updateMercury( completeness ) {
         const trackHeight = this.widget.querySelector( '.range-widget__ticks' ).clientHeight;
         const bulbHeight = this.widget.querySelector( '.range-widget__bulb' ).clientHeight;
         this.widget.querySelector( '.range-widget__bulb__mercury' ).style.height = `${( completeness * trackHeight ) + ( 0.5 * bulbHeight )}px`;
     }
 
+    /**
+     * @param {object} props
+     * @return {string} HTML string
+     */
     _stepsBetweenHtmlStr( props ) {
         let html = '';
         if ( props.distress ) {
@@ -118,6 +129,9 @@ class RangeWidget extends Widget {
         return html;
     }
 
+    /**
+     * Resets widget
+     */
     _reset() {
         // Update UI stuff before the actual value to avoid issues in custom clients that may want to programmatically undo a reset ("strict required" in OpenClinica)
         // as that is subtly different from updating a value with a calculation since this.originalInputValue=  sets the evaluation cascade in motion.
@@ -127,14 +141,23 @@ class RangeWidget extends Widget {
         this.originalInputValue = '';
     }
 
+    /**
+     * Disables widget
+     */
     disable() {
         this.widget.querySelectorAll( 'input, button' ).forEach( el => el.disabled = true );
     }
 
+    /**
+     * Enables widget
+     */
     enable() {
         this.widget.querySelectorAll( 'input, button' ).forEach( el => el.disabled = false );
     }
 
+    /**
+     * Updates widget
+     */
     update() {
         const value = this.element.value;
 
@@ -146,6 +169,9 @@ class RangeWidget extends Widget {
         }
     }
 
+    /**
+     * @type object
+     */
     get props() {
         const props = this._props;
         const min = isNumber( this.element.getAttribute( 'min' ) ) ? this.element.getAttribute( 'min' ) : 0;
@@ -164,6 +190,9 @@ class RangeWidget extends Widget {
         return props;
     }
 
+    /**
+     * @type string
+     */
     get value() {
         return this.range.classList.contains( 'empty' ) ? '' : this.range.value;
     }
--- a/src/widget/rating/rating.js
+++ b/src/widget/rating/rating.js
@@ -1,0 +1,134 @@
+import Widget from '../../js/widget';
+import { isNumber } from '../../js/utils';
+import events from '../../js/event';
+
+/**
+ * @extends RangeWidget
+ */
+class RatingWidget extends Widget {
+    /**
+     * @type string
+     */
+    static get selector() {
+        return '.or-appearance-rating input[type="number"]';
+    }
+
+    _init() {
+        const fragment = document.createRange().createContextualFragment( this._getHtmlStr() );
+        this.element.after( fragment );
+        this.element.classList.add( 'hide' );
+        this.widget = this.question.querySelector( '.widget' );
+
+        let ratingStars = '',
+            name = Math.random().toString( 36 ).substring( 2, 15 );
+        for ( let i = this.props.min; i <= this.props.max; i += this.props.step ) {
+            ratingStars += `<input type=radio name="rating-stars__${name}" class="rating-widget__rating__star ignore" value="${i}"/>`;
+        }
+
+        this.widget.querySelector( '.rating-widget__rating' )
+            .append( document.createRange().createContextualFragment( ratingStars ) );
+
+        this.rating = this.widget.querySelector( '.rating-widget__rating' );
+        this.stars = this.rating.querySelectorAll( '.rating-widget__rating__star' );
+        this.stars.forEach( el => {
+            el.addEventListener( 'change', () => {
+                const selected = this.rating.querySelector( 'input:checked' );
+                // contains class 'empty' means first load
+                if ( ( this.value != selected.value || this.rating.classList.contains( 'empty' ) ) ) {
+                    this.value = selected.value;
+                }
+            } );
+        } );
+
+        // loads the default value if exists
+        this.update();
+
+        if ( this.props.readonly ) {
+            this.disable();
+        }
+    }
+
+    /**
+     * This is separated so it can be extended (in the analog-scale widget)
+     *
+     * @return {string} HTML string
+     */
+    _getHtmlStr() {
+        return `<div class="widget rating-widget">
+                	<div class="rating-widget__rating empty"></div>
+            	</div>`;
+    }
+
+    /**
+     * Disables widget
+     */
+    disable() {
+        this.stars.forEach( el => el.disabled = true );
+    }
+
+    /**
+     * Enables widget
+     */
+    enable() {
+        this.stars.forEach( el => el.disabled = false );
+    }
+
+    /**
+     * Updates widget
+     */
+    update() {
+        const value = this.element.value;
+        if ( isNumber( value ) ) {
+            this.stars.forEach( ( star ) => {
+                if ( star.value === value ) {
+                    star.checked = true;
+                    star.dispatchEvent( events.Change() );
+                }
+            } );
+        } else {
+            this._reset();
+        }
+    }
+
+    /**
+     * Resets widget
+     */
+    _reset() {
+        const selected = this.rating.querySelector( 'input:checked' );
+        if ( selected ) {
+            this.value = '';
+            selected.checked = false;
+        }
+    }
+
+    /**
+     * @type object
+     */
+    get props() {
+        const props = this._props;
+        const min = isNumber( this.element.getAttribute( 'min' ) ) ? this.element.getAttribute( 'min' ) : 0;
+        const max = isNumber( this.element.getAttribute( 'max' ) ) ? this.element.getAttribute( 'max' ) : 10;
+        const step = isNumber( this.element.getAttribute( 'step' ) ) ? this.element.getAttribute( 'step' ) : 1;
+
+        props.min = Number( min );
+        props.max = Number( max );
+        props.step = Number( step );
+
+        return props;
+    }
+
+    /**
+     * @type string
+     */
+    get value() {
+        return this.originalInputValue;
+    }
+
+    set value( value ) {
+        this.originalInputValue = value;
+        this.rating.classList.toggle( 'empty', value === '' );
+    }
+
+}
+
+export default RatingWidget;
--- a/src/widget/rating/rating.scss
+++ b/src/widget/rating/rating.scss
@@ -1,0 +1,39 @@
+.or-appearance-rating {
+
+    .rating-widget__rating {
+        display: inline-block;
+        width: 100%;
+        height: auto;
+    }
+
+    input[type=radio] {
+        &.rating-widget__rating__star, &.rating-widget__rating__star:checked {
+            appearance: none;
+            -moz-appearance: none;
+            -webkit-appearance: none;
+            -ms-appearance: none;
+            display: inline-block;
+            width: 40px;
+            height: 40px;
+            margin-right: 5px;
+            margin-bottom: 0;
+            margin-top: 0;
+            border-style: solid;
+            background-color: transparent;
+            background-image: svg-as-url("<svg xmlns='http://www.w3.org/2000/svg' viewBox='-6 1 53 50'> 0 3 <path d='m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z' style='fill:#{$brand-primary-color}' /> 0 4 </svg>");
+            background-size: 35px 25px;
+            background-position: -3px;
+            background-repeat: no-repeat;
+            border-color: transparent !important;
+            border-radius: 0;
+        }
+    }
+
+    /* plain stars, hover behavior */
+    .empty input.rating-widget__rating__star:not(:hover),
+    .empty input.rating-widget__rating__star:hover:disabled,
+    .rating-widget__rating__star:checked ~ .rating-widget__rating__star:not(:hover),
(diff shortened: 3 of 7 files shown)


theywa:
> I think it might be intuitive to unselect (reset) by clicking the star. Not sure if we'd need to keep the reset button if we have that. What do you think?

Yes, we don't need the reset button anymore if it implemented that way.

> Loading a default value seems to not work correctly (e.g. if expect 3 stars, it shows 2)

I think we need set what is star represents and its value because it's work as expected by manually set the value on the update method? 
currently, I set tick as a star (including the value of its star based on tick value).
Should I use the value based on the order(like the first star mean 1, second = 2, etc)?

> The widget doesn't work with keyboard tabbing, selecting, and doesn't show focus.

Yes, you're right. It just spans so tabbing and focus doesn't work. Thanks for the clue and example, I will update the code.

MartijnR:
> currently, I set tick as a star (including the value of its star based on tick value).
> Should I use the value based on the order(like the first star mean 1, second = 2, etc)?

If you're using the `<input>`s from the codepen, you could use the semantically correct `value` attribute. So when you load a value, you'll find the input with that value and set `input.checked=true`. Note, that you could potentially support min=3, max=7, step=1, or even step=3, so you'll want to take that into account, so best to explicitly set values for each star.

theywa:
@MartijnR - sorry for the late response, I update the code but still in same widget file.
- Use radio as a star
- Use minimal js action
- Make sure tab and focus working
Preview
Screenshot from 2019-09-08 20-10-39

src/widget/range/range-widget.scss
```diff
@@ -389,3 +389,56 @@ $green: rgb(60, 182, 67);
...
+                    background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='-6 1 53 50'%3E%0A03 %3Cpath d='m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z' style='fill:%23f76411;'/%3E%0A04%0A%3C/svg%3E");
+                }
+            }
+        }
+    }
+
+    /* plain stars, hover behavior */
+    .empty input.star:not(:hover),
+    .star:checked ~ .star:not(:hover) {
+        border-color: #ffffff;
```
MartijnR: border-color should be transparent, because repeats have a different background and so do some themes when hovering over the question (and a theme could also provide a custom background for the whole form). Alternatively, and I like this better, is to use padding, and `background-repeat: no-repeat;`, and positioning the background images.

test/spec/widget.range.spec.js
```diff
@@ -9,6 +9,11 @@ const FORM1 =
...
     </label>`;
 const FORM2 = FORM1.replace( 'step="1"', 'step="0.1"' );
+const FORM3 =
+    `<label class="question non-select or-appearance-rating">
+        <span lang="" class="question-label active">Range</span>
+        <input type="number" name="/widgets/range_widgets/range3" data-type-xml="int" min="0" max="5" step="1">
+    </label>`;
 
 runAllCommonWidgetTests( RangeWidget, FORM1, '2' );
 runAllCommonWidgetTests( RangeWidget, FORM2, '2.1' );
```
MartijnR: Try running the commonWidgetTest for your template/widget combo

theywa:
@MartijnR - Got it, but did this ticket have a deadline or is it possible to continue it later(at most 10 days later)?

MartijnR:
Take your time! A few weeks is fine.

src/widget/range/range-widget.scss
```diff
@@ -389,3 +389,56 @@ $green: rgb(60, 182, 67);
...
+                height: 40px;
+                margin-right: 5px;
+                margin-bottom: 0;
+                margin-top: 0;
+                border-width: 10px;
+                border-style: solid;
+                background-color: transparent;
+                background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='-6 1 53 50'%3E%0A03 %3Cpath d='m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z' style='fill:%23f76411;'/%3E%0A04%0A%3C/svg%3E");
+                background-size: 20px 20px;
+                background-position: -3px;
```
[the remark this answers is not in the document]
theywa: I remove the border-width, and it looks centered now. Let me know if it still not centered enough.

src/widget/range/range-widget.js
```diff
@@ -15,49 +15,55 @@ class RangeWidget extends Widget {
...
+            } );
+
+            this.widget.querySelector( 'input.empty' ).addEventListener( 'touchstart', () => {
+                this.range.classList.remove( 'empty' );
+                this.range.dispatchEvent( events.Change() );
+            } );
+        } else {
+            this.element.after( fragment );
+            this.widget = this.question.querySelector( '.widget' );
+            this.range = that.element;
```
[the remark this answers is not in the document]
theywa: yes my mistake, code updated

src/widget/range/range-widget.js
```diff
@@ -67,17 +73,47 @@ class RangeWidget extends Widget {
...
+                // contains class 'empty' means first load
+                if ( selected && ( this.originalInputValue != selected.value || this.widgetThick.classList.contains( 'empty' ) ) ) {
+                    this.value = selected.value;
+                } else {
+                    this.value = null;
+                    selected.checked = false;
+                }
+                this.widgetThick.classList.toggle( 'empty', !this.value );
+                this.originalInputValue = this.value;
+                this.range.value = this.value;
```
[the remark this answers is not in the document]
theywa: my mistake

src/widget/range/range-widget.scss
```diff
@@ -389,3 +389,56 @@ $green: rgb(60, 182, 67);
...
+                display: inline-block;
+                width: 40px;
+                height: 40px;
+                margin-right: 5px;
+                margin-bottom: 0;
+                margin-top: 0;
+                border-width: 10px;
+                border-style: solid;
+                background-color: transparent;
+                background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='-6 1 53 50'%3E%0A03 %3Cpath d='m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z' style='fill:%23f76411;'/%3E%0A04%0A%3C/svg%3E");
```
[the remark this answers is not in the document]
theywa: code updated

src/widget/range/range-widget.scss
```diff
@@ -389,3 +389,56 @@ $green: rgb(60, 182, 67);
...
+                background-size: 20px 20px;
+                background-position: -3px;
+                border-color: #ffffff;
+                border-radius: 0;
+
+                &:checked {
+                    border-color: #ffffff;
+                    background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='-6 1 53 50'%3E%0A03 %3Cpath d='m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z' style='fill:%23f76411;'/%3E%0A04%0A%3C/svg%3E");
```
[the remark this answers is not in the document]
theywa: you're right, duplicate code removed

src/widget/range/range-widget.js
```diff
@@ -67,17 +73,47 @@ class RangeWidget extends Widget {
...
+
         // Various attemps to use more elegant CSS background on the _ticks div, have failed due to little
         // issues seemingly related to rounding or browser sloppiness. This far is less elegant but nice and robust:
         this.widget.querySelector( '.range-widget__ticks' )
-            .append( document.createRange().createContextualFragment( new Array( ticks ).fill( '<span></span>' ).join( '' ) ) );
+            .append( document.createRange().createContextualFragment( addTicks ) );
+
+        if ( this.props.rating ) {
+            this.widgetThick = this.widget.querySelector( '.range-widget__ticks' );
+            this.widgetThick.addEventListener( 'click', () => {
```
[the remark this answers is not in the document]
theywa: since the current rating doesn't have any reset button so I use onClick to toggle the selected button.

src/widget/range/range-widget.js
```diff
@@ -15,49 +15,55 @@ class RangeWidget extends Widget {
...
 
-        this.widget.querySelector( '.btn-reset' ).addEventListener( 'click', this._reset.bind( this ) );
+            this.widget = this.question.querySelector( '.widget' );
+            this.range = this.widget.querySelector( 'input' );
+            this.current = this.widget.querySelector( '.range-widget__current' );
 
-        // loads the default value if exists, else resets
-        this.update();
+            if ( this.props.readonly ) {
+                this.disable();
```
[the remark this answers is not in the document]
theywa: code updated

src/widget/range/range-widget.scss
```diff
@@ -389,3 +389,56 @@ $green: rgb(60, 182, 67);
...
+.or-appearance-rating {
+
+    .rating {
+        display: inline-block;
+        width: 100%;
+        height: auto;
+    }
+
+    input {
+        &:not([type=radio]) {
```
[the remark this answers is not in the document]
theywa: yes, code updated

src/widget/range/range-widget.scss
theywa: code updated

test/spec/widget.range.spec.js
theywa: tried that but it's not working since a lot of doms is missing in the rating widget. Do you have any suggestions what kind of test unit should I create?

theywa:
@MartijnR - Thanks, I'm back. Just update the PR with the new widget can you take a look? will try to add the test unit tomorrow.

src/widget/rating/rating.js
```diff
@@ -0,0 +1,117 @@
+import RangeWidget from '../../widget/range/range-widget';
+import { isNumber } from '../../js/utils';
+
+/**
+ * @extends RangeWidget
+ */
+class RatingWidget extends RangeWidget {
```
MartijnR: Are we using any of the functions in the RangeWidget? If not, we should extend Widget instead.

src/widget/rating/rating.js
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    /**
+     * This is separated so it can be extended (in the analog-scale widget)
+     *
+     * @return {string} HTML string
+     */
+    _getHtmlStr() {
+        return `<div class="widget range-widget">
```
MartijnR: would `class="widget rating-widget"` work?

MartijnR: Would be nice to eliminate all references in css and javascript (variables) to "ticks" and "range-*"

src/widget/rating/rating.js
MartijnR: I see you are using `get props()`. Makes sense.

src/widget/rating/rating.js
```diff
@@ -0,0 +1,117 @@
...
+        return '.or-appearance-rating input[type="number"]';
+    }
+
+    _init() {
+        const fragment = document.createRange().createContextualFragment( this._getHtmlStr() );
+        this.element.after( fragment );
+        this.element.classList.add( 'hide' );
+        this.widget = this.question.querySelector( '.widget' );
+
+        let ticks = this.props.ticks ? Math.ceil( Math.abs( ( this.props.max - this.props.min ) / this.props.step ) ) : 1;
```
MartijnR: this.props.ticks is not relevant for the rating widget, so that check can be removed. Stars will always be shown.

src/widget/rating/rating.js
```diff
@@ -0,0 +1,117 @@
...
+        // Now reduce to a number < 50 to avoid showing a sold black tick line.
+        let divisor = Math.ceil( ticks / this.props.maxTicks );
+        while ( ticks % divisor && divisor < ticks ) {
+            divisor++;
+        }
+        ticks = ticks / divisor;
+
+        let addTicks = new Array( ticks ).fill( '<span></span>' ).join( '' );
```
MartijnR: Looks to me that the addTicks variable is not required for this widget

src/widget/rating/rating.js
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * This is separated so it can be extended (in the analog-scale widget)
+     *
+     * @return {string} HTML string
+     */
+    _getHtmlStr() {
+        return `<div class="widget range-widget">
+                	<div class="range-widget__ticks rating empty"></div>
```
MartijnR: `class="rating-widget__rating empty"` perhaps?

src/widget/rating/rating.js
```diff
@@ -0,0 +1,117 @@
...
+            name = Math.random().toString( 36 ).substring( 2, 15 );
+        for ( let i = this.props.min; i <= this.props.max; i += this.props.step ) {
+            ratingStars += `<input type=radio name="rating-stars__${name}" class="star rating__tick" value="${i}"/>`;
+        }
+        addTicks = ratingStars;
+
+        this.widget.querySelector( '.range-widget__ticks' )
+            .append( document.createRange().createContextualFragment( addTicks ) );
+
+        this.widgetThick = this.widget.querySelector( '.range-widget__ticks' );
```
MartijnR: `this.rating = this.widget.querySelector('.range-widget__rating')` perhaps

src/widget/rating/rating.js
```diff
@@ -0,0 +1,117 @@
...
+        while ( ticks % divisor && divisor < ticks ) {
+            divisor++;
+        }
+        ticks = ticks / divisor;
+
+        let addTicks = new Array( ticks ).fill( '<span></span>' ).join( '' );
+        let ratingStars = '',
+            name = Math.random().toString( 36 ).substring( 2, 15 );
+        for ( let i = this.props.min; i <= this.props.max; i += this.props.step ) {
+            ratingStars += `<input type=radio name="rating-stars__${name}" class="star rating__tick" value="${i}"/>`;
```
MartijnR: `class="star"` or `class="rating-widget__rating__star"` would be nicer perhaps

src/widget/rating/rating.scss
```diff
@@ -0,0 +1,41 @@
...
+                border-radius: 0;
+            }
+        }
+    }
+
+    /* plain stars, hover behavior */
+    .empty input.star:not(:hover),
+    .star:checked ~ .star:not(:hover),
+    .star:checked ~ .star:hover:disabled {
+        background-image: svg-as-url("<svg xmlns='http://www.w3.org/2000/svg' viewBox='-6 1 53 50'> 0 3 <path d='m25,1 6,17h18l-14,11 5,17-15-10-15,10 5-17-14-11h18z' /> 0 4 </svg>");
```
MartijnR: without fill to avoid confusion with selected stars

src/widget/rating/rating.js
```diff
@@ -0,0 +1,117 @@
...
+        }
+        addTicks = ratingStars;
+
+        this.widget.querySelector( '.range-widget__ticks' )
+            .append( document.createRange().createContextualFragment( addTicks ) );
+
+        this.widgetThick = this.widget.querySelector( '.range-widget__ticks' );
+        this.widgetThick.querySelectorAll( '.star.rating__tick' ).forEach( el => {
+            el.addEventListener( 'click', () => {
```
MartijnR: Since these are `input` elements, I think we should use `change` or `input`. This `click` listener does seem to work, but i have a suspicion we may be missing some special case/device/browser issue.

test/spec/widget.range.spec.js
MartijnR: For now, it would be fine to just runAllCommonWidgetTests() for a few combinations of step, min, max (as in `widget.range.spec.js`). No need to add custom tests, I think. So the spec file should be just a few lines of code. If tests are failing for unknown reasons, just add the spec file to this PR and I can help. It might be a yet undiscovered bug in the widget (that's why these common tests are so fabulous).

src/widget/rating/rating.js
theywa: changed to extend Widget class directly by adding it's own `get props()`

src/widget/rating/rating.js
theywa: code updated

src/widget/rating/rating.js
theywa: code removed

src/widget/rating/rating.js
theywa: code removed

src/widget/rating/rating.js
theywa: code updated

src/widget/rating/rating.js
theywa: code updated

src/widget/rating/rating.js
theywa: code updated

src/widget/rating/rating.scss
theywa: style updated

src/widget/rating/rating.js
theywa: code updated

theywa:
@MartijnR - code updated add `_reset()` to fixed issue with resetting calculate value. Also, no error when  `runAllCommonWidgetTests()` . A little bit OOT, what is `.ignore` used for? I don't see extra style attached to that class.

test/spec/widget.range.spec.js
theywa: code updated and no error found after run the test

src/widget/rating/rating.js
MartijnR: Nice!

MartijnR:
Great! Thanks for your persistence!

The `ignore` class is used outside of the widget in the form engine. The form engine monitors all form controls for changes except the ones with the `ignore` class. 

For the rating widget it monitors the original `<input type="number">` that you are instantiating your widget on. The widget is a layer on top of those original form controls. Any form controls (`<input>`, `<select>`, `<textarea>`) that the widget adds (like your star `<input>`s), should only be handled by the widget itself - and the widget has to change the original form control when needed and fire a change event on it. If you don't add the ignore class to those added inputs, the form engine is going to try to go through the engine stuff whenever a star is clicked but fails (you'll see a warning message in the console) because those inputs don't have all the additional info (which date node it is linked to, the logic etc).

This is ready to be merged. If you have to time to move the tests to a separate widget.rating.spec.js file (no problem to duplicate HTML templates for tests), please do so (in that case, you could also remove the .ignore part from the various querySelectors for further perfection). Otherwise, no problem, and I'll merge as is on Monday and do that myself.

theywa:
No problem, cleanup code is done.

MartijnR:
👍

Issue #590, opened 2018-12-19
Advanced hierarchical Table of Contents


MartijnR:
To be considered. Like ODK Collect showing hierarchy. Also see ongoing discussion: https://forum.opendatakit.org/t/new-xform-appearance-hierarchy-visible-groups/16858/2

TL/DR; wrt groups, only show a group as a group in the hierarchy if it has a label

MartijnR:
* [ ] create a complex test form
* [ ] see how it works in Collect
* [ ] mimic this if that looks good

gushil:
Hi @MartijnR , need some clarification.

From what I read in discussion and by using ODK Collect to open forms mentioned in https://github.com/opendatakit/collect/pull/2764, these are targets for this issue:

- [ ] Show TOC if there is a group with a label in a form, even when the form is only one page.
- [ ] Show Nested Group as nested label.
- [ ] Group TOC item is different (ODK Collect using some 'folder' icon) with Page TOC item.
- [ ] Click Group TOC item made us jump into the right form item.

MartijnR:
@gushil, good questions. We can deviate a little from ODK Collect because we have pages and non-pages mode (and Collect is always in pages mode).

1. Enketo will only show a ToC if a form is in pages mode (and yes, if it has one page, it will be shown). So this can remain as it is currently.
2. Groups are only shown in the ToC if they have a label (I believe this is what Collect does?). Currently, Enketo doesn't show groups at all.
3. All questions are shown (if that's the case in ODK Collect). Currently, Enketo only shows one item per page.
3. Clicking any item in the ToC (group or question) jumps to that item

Could you share your test form in this thread please? (Google Drive ideally)

gushil:
I'm using this test form https://docs.google.com/spreadsheets/d/1UD7QxArWhgJOgC18uQANupa4TAmZbWKOEdaqSmkdN0o/edit?usp=sharing

gushil:
It looks like groups already shown in ToC ?

![image](https://user-images.githubusercontent.com/60411/64164451-7dde7980-ce6d-11e9-96a4-8d006949acce.png)

MartijnR:
Hi Agus,

No, it doesn't include groups in Enketo's current implementation, I think. It just extracts the first question label (or hint) from each page and shows that in the ToC.

[...]

PR #660, opened 2019-09-05
fixed: Advanced hierarchical Table of Contents

gushil:
#590


gushil:
Fixing some test scripts.

src/js/page.js
```diff
@@ -20,6 +20,7 @@ export default {
         if ( this.form.view.$.hasClass( 'pages' ) ) {
             const $allPages = this.form.view.$.find( ' .question:not([role="comment"]), .or-appearance-field-list' )
                 .add( '.or-repeat.or-appearance-field-list + .or-repeat-info' )
+                .add( '.or-group' )
```
MartijnR: I think this turns all groups into a page. That's not right. I don't think this needs any change. There are no known bugs in how a form is divided into pages, so that should not changes. This feature is purely ToC related.

src/js/page.js
```diff
@@ -336,14 +338,44 @@ export default {
...
-            a.setAttribute( 'href', `#${item.pageEl.querySelector( '[name]' ).getAttribute( 'name' )}` );
-            a.textContent = item.tocItemText;
-            li.append( a );
-            items.appendChild( li );
+            if ( $itemPageEl.is( '.or-group' ) ) {
+                const details = document.createElement( 'details' );
+                const summary = document.createElement( 'summary' );
+                const groupList = document.createElement( 'ul' );
+                summary.textContent = item.tocItemText;
+                $( details ).addClass( 'groups-toc' );
```
MartijnR: we're slowly removing jQuery, so let's use `details.classList.add('groups-toc')`.

src/js/page.js
```diff
@@ -336,14 +338,44 @@ export default {
     },
     _getTocHtmlFragment( tocItems ) {
         const items = document.createDocumentFragment();
+        let currentGroupList;
         tocItems.forEach( item => {
+            const $itemPageEl = $( item.pageEl );
```
MartijnR: let's not use jQuery. We're in the process of removing it, and it has no advantage here.

src/js/page.js
```diff
@@ -336,14 +338,44 @@ export default {
...
+                currentGroupList = groupList;
+                li.append( details );
+                items.appendChild( li );
+            } else {
+                const a = document.createElement( 'a' );
+                a.setAttribute( 'href', `#${item.pageEl.querySelector( '[name]' ).getAttribute( 'name' )}` );
+                a.textContent = item.tocItemText;
+                li.setAttribute( 'role', 'pageLink' );
+                li.append( a );
+                if ( $itemPageEl.parent().is( '.or-group' ) ) {
```
MartijnR: `item.pageEl.matches('.or-group')`

gushil:
image

gushil:
image

gushil:
image

gushil:
image

gushil:
@MartijnR 

Thanks for the response.

```
- a
> group one
    - b
    - c
- d
- **e**
> group three
    - **f**
    - **g**
    > **group three-two**
        - **h**
        - **i**
```

However, I still not understand why items in ** will be shown in ToC, because if we navigate form with buttons (Next and Back buttons), those items will not shown as its own page.

MartijnR:
Hi @gushil, The way I understood it ToC content and structure has basically nothing to do with pages. But I'm going to check with ODK Collect, because I might be misunderstanding!

src/js/page.js
```diff
@@ -162,8 +162,8 @@ export default {
         this.$toc
             .on( 'click', 'a', function() {
                 if ( !that.form.pageNavigationBlocked ) {
-                    const index = $( this.parentNode ).prevAll().length;
-                    that.flipToPageContaining( $( that.tocItems[ index ].pageEl ) );
+                    const currentIndex = that.$toc.find( '[role="pageLink"]' ).index( $( this.parentNode ) );
+                    that.flipToPageContaining( $( that.tocItems[ currentIndex ].pageEl ) );
```
MartijnR: Sorry, I didn't notice this before. This only flips to the page containing the ToC element. We can perhaps do even better than that by using that.form.goToTarget(questionOrGroup). You may have to build a map of tocItem -> element.

MartijnR:
Yes, this is correct. I understand the confusion though. Just having the ToC code as part of pages.js is confusing. This was actually the case for historical reasons (since the current ToC **does** relate to pages). I think it is worth considering creating toc.js maybe, and importing that in either form.js (best?) or pages.js. Please consider that, and use the same toc-update trigger as pages.js uses to update pages.

gushil:
Thanks for the suggestions!
Will look into that possibility and implementing it next.

gushil:
image

Almost there. Need to refactor and do some styling.

gushil:
image

PR #663, opened 2019-09-12
[DNM] changed: first try replacing the xpath-evaluator

MartijnR:
Hi @vimemo,

I got a little excited and wanted to see how your new evaluator worked in Enketo Core.

There are some failing tests in Enketo Core. From a first glance it looks like most (or all) are related to XPathResult data types. And maybe some to some more complex internal XPath queries. I'll look further, but I wanted to quickly share this already to see if you had the same idea about the method of replacing enketo-xpathjs (and maybe you've already tried this yourself).

It's a bit messy due to the whitespace changes, but the main change is in xpath-evaluator-binding.js. Does that way of binding seem correct? (not urgent)

I also understood that we can no longer rename the custom evaluate function (to e.g. jsEvaluate as we had before). I think that's fine if the new evaluator is nearly as fast for purely native XPath expressions. It will simplify the code in enketo-core.

Cheers,

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8123,6 +8123,11 @@
                 "lodash.toarray": "^4.4.0"
             }
         },
+        "node-forge": {
+            "version": "0.9.0",
+            "resolved": "https://registry.npmjs.org/node-forge/-/node-forge-0.9.0.tgz",
+            "integrity": "sha512-7ASaDa3pD+lJ3WvXFsxekJQelBKRpne+GOVbLbtHYdd7pFspyeuJHnWfLplGf3SwKGbfs/aYl5V/JCIaHVUKKQ=="
+        },
         "node-gyp": {
             "version": "3.8.0",
             "resolved": "https://registry.npmjs.org/node-gyp/-/node-gyp-3.8.0.tgz",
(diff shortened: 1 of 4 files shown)


MartijnR:
WIP: 

Looks like the (main) issue is that when requesting result type 7 (ORDERED_NODE_SNAPSHOT_TYPE), the actual result is type 4, and not a snapshot. It doesn't have the `snapshotLength` property.

MartijnR:
It looks like those tests (e.g. with `numberValue` and `snapshotLength`) are there in enketo-xpathjs and pass after replacing with your evaluator. So I'm puzzled about this. We're binding the evaluator to an XML document instead of the window document here. Could the issue be related to how binding to an XML Doc works in `openrosa-xpath-evaluator`?

MartijnR:
Hi @vimemo,

Just wanted to ping you to see if you have time to check how your XPath should be added to enketo-core. Trying to get most of the work done before the end of the month and start the final report.

vimemo:
hi @MartijnR, thanks for your patience on this. Will try to get it done this week. should get back to you by this coming Monday at the latest.

MartijnR:
That's awesome. Thank you!

Issue #665, opened 2019-09-24
Time question with value between 12:00 with 13:00 is not able to render the correct 12PM time .


nanzhang01:
This issue impacts both time and datetime questions.
Sample xls form:
[sample for time question issue.xlsx](https://github.com/enketo/enketo-core/files/3645538/sample.for.time.question.issue.xlsx)
![image](https://user-images.githubusercontent.com/29697780/65486334-c3272180-ded6-11e9-9052-cf38caf0241b.png)

preview after loaded:
![image](https://user-images.githubusercontent.com/29697780/65486365-d4702e00-ded6-11e9-9df5-d48d77fa3e3c.png)

nanzhang01:
@MartijnR , apologies, this fix introduce new bug: time value between 00:00-01:00 will be converted to pm if `showMeridian` is true after clicking the time picker.
![image](https://user-images.githubusercontent.com/29697780/69072422-93c00b80-0a66-11ea-9b52-df09728f30dd.png)
![error3](https://user-images.githubusercontent.com/29697780/69072706-1cd74280-0a67-11ea-81b7-2c206e234051.gif)

Issue #666, opened 2019-09-24
Time will always change to AM when 12 hours mode is enabled in non-english language.


nanzhang01:
Steps to reproduce:
-  Set the env language as non-english, eg: zh-cn.
-  Set the time value between 12:00-23:59. 
-  Click the time picker widget, observe that the question will change to AM.
- Click the time question again, change the time from AM to PM, finish the question and observe that it stays in PM, but if click the question again it will change to AM.

![timebugs](https://user-images.githubusercontent.com/29697780/65499460-37ba8a00-def0-11e9-8c96-90847467e186.gif)

Issue #649, opened 2019-08-14
datetime question with midnight datetime default does not display time component


MartijnR:
![62851435-a5cc3880-bd18-11e9-99d8-f31d16ebc6c6](https://user-images.githubusercontent.com/627350/63056725-e0171080-bea5-11e9-934a-191fdb71942e.png)
![62851334-57b73500-bd18-11e9-9fb4-27c5683c10d5](https://user-images.githubusercontent.com/627350/63056748-eb6a3c00-bea5-11e9-9135-14f125aa3958.png)

MartijnR:
After looking closer at this, the first of these does not appear to be a bug, or is a very different bug. I think the correct behavior would be to not recognize this value and show empty date and time (since the date format is not supported). Similarly, I think `"2016-10-08"` should be considered and invalid datetime.

nanzhang01:
Regression happens for this issue. The current behaviour is like:
![image](https://user-images.githubusercontent.com/29697780/69065142-37ef8580-0a5a-11ea-83e2-b839182d7a4d.png).
This regression is introduced from https://github.com/enketo/enketo-core/commit/896e0b9a45c6e900811791447983b709bb07f306#diff-fb12358d88df2a187c353c9812b4014eR257,  `new Date('2019-11-01T00:00:00+08:00').toISOLocalString()` only returns the date part,  seems it's  by design result defined in enketo-xpathjs.

PR #667, opened 2019-09-25
Fix three bugs on time picker and datetime widget

nanzhang01:
Fix:
#665 
#649 
#666

### DIFF
--- a/src/widget/datetime/datetimepicker-extended.js
+++ b/src/widget/datetime/datetimepicker-extended.js
@@ -168,7 +168,11 @@ class DatetimepickerExtended extends Widget {
         const val = value ? new Date( value ).toISOLocalString() : '';
         const vals = val.split( 'T' );
         const dateVal = vals[ 0 ];
-        const timeVal = ( vals[ 1 ] && vals[ 1 ].length > 4 ) ? vals[ 1 ].substring( 0, 5 ) : '';
+        /**
+         * seems the source of issue #649 is in the toISOLocalString function 
+         * refer: https://github.com/enketo/enketo-xpathjs/blob/master/src/date-extensions.js#L16
+         */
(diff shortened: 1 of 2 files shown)


src/widget/datetime/datetimepicker-extended.js
```diff
@@ -168,7 +168,11 @@ class DatetimepickerExtended extends Widget {
         const val = value ? new Date( value ).toISOLocalString() : '';
         const vals = val.split( 'T' );
         const dateVal = vals[ 0 ];
-        const timeVal = ( vals[ 1 ] && vals[ 1 ].length > 4 ) ? vals[ 1 ].substring( 0, 5 ) : '';
+        /**
+         * seems the source of issue #649 is in the toISOLocalString function 
+         * refer: https://github.com/enketo/enketo-xpathjs/blob/master/src/date-extensions.js#L16
```
MartijnR: Yes, I think you're right. That is not a great trick (in date-extensions.js) but I don't think we can remove it atm.

MartijnR: It was introduced here: https://github.com/enketo/enketo-xpathjs/commit/5dbfa2107a9ca9efe072608bc1201b1c927d37bf#diff-eb47f20ec88abc06d935a1c5511de69f

src/widget/datetime/datetimepicker-extended.js
```diff
@@ -168,7 +168,11 @@ class DatetimepickerExtended extends Widget {
...
+         */
+        const timeVal = ( vals[ 1 ] && vals[ 1 ].length > 4 ) ? vals[ 1 ].substring( 0, 5 ) : ( dateVal && !vals[ 1 ] ) ? '00:00' : '';
```
MartijnR: I don't think this workaround will cause any harm, so looks good to me, and thanks for adding that comment!

MartijnR:
Thanks a lot!

PR #684, opened 2019-11-15
Print view does not display long text in its entirety

### LINKED ISSUE #166: Print view does not display long text in its entirety
The text gets cut off in printout. 

To reproduce: 

1. load test form: https://docs.google.com/spreadsheets/d/1qHG7CJyhb2xzyTMzzq03iyDeL_JXRlLsSpKEmwBICUs/edit?usp=sharing
2. print (or view print preview) 

There are two different issues:

1. Simple text input
2. Multiline text input

in two different scenarios:

a. default theme (formhub)
b. grid theme

Would be great if there was a css-only solution, but it probably needs some javascript magic.

There is a js print module that is called when a user in Enketo Express clicks the print button (and NOT the print command in the browser directly). This function is not exposed in enketo-core, but can be called be called e.g. by adding a few lines in app.js or via the console. See https://github.com/enketo/enketo-core/blob/master/src/js/print.js.

gushil:
fix #166

### DIFF
--- a/src/widget/text-print/text-print.js
+++ b/src/widget/text-print/text-print.js
@@ -1,0 +1,45 @@
+import Widget from '../../js/widget';
+
+/**
+ * Clone text input fields value into new print-only element.
+ * This is an unusual way to implement a feature, because it is not an actual widget,
+ * but this is the easiest way to do it.
+ */
+class TextPrintWidget extends Widget {
+    /**
+     * @type string
+     */
+    static get selector() {
+        return 'input[type=text][data-type-xml="string"], textarea[data-type-xml="string"]';
+    }
+
+    _init() {
+        const printElement = document.createElement( 'div' );
+
+        const elementTagName = this.element.tagName.toLowerCase();
+        let printElementClassName = 'print-' + elementTagName;
+        if ( elementTagName === 'input' ) {
+            printElementClassName += '-' + this.element.type;
+        }
+        printElement.className = printElementClassName + ' print-only';
+
+        if ( getComputedStyle( this.element ).order !== '' ) {
+            printElement.style.order = parseInt( getComputedStyle( this.element ).order, 10 ) + 1;
+        }
+        this.element.after( printElement );
+
+        const widget = this.element.parentElement.querySelector( '.' + printElementClassName );
+        if ( this.element.value !== '' ) {
+            const widgetText = this.element.value.replace( /\n/g, '<br>' );
+            widget.innerHTML = widgetText;
+        }
+
+        this.element.addEventListener( 'input', event => {
+            const el = event.target;
+            const widgetText = el.value.replace( /\n/g, '<br>' );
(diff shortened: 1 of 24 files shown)


gushil:
Sorry for not addressing the feedbacks soon. Very impressed with the quality of your feedback. Learning a ton!

Quick question: Already moved all styles to widget folder. How do I style widget a little bit different for grid-theme?

image

There is still border-bottom on input text. I want to remove that.

src/widget/text-print/text-print.js
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    update() {
+        if ( this.element.getAttribute( 'data-calculate' ) !== null && this.element.getAttribute( 'data-calculate' ) !== '' ) {
+            const dataCalculateValue = this.element.getAttribute( 'data-calculate' ).trim();
+            const dataCalculateElement = document.querySelector( `[name="${dataCalculateValue}"]` );
+            if ( dataCalculateElement ) {
+                const widget = this.element.parentElement.querySelector( `.${this._getClassName()}` );
+                if ( dataCalculateElement.value !== '' ) {
+                    const widgetText = dataCalculateElement.value.replace( /\n/g, '<br>' );
```
MartijnR: Repeated code (3x). Create a function, e.g. `_copyValue` and maybe make this.widget a class instance property (element) during init so it will only querySelect once.


So the update function may only become:

```js
update(){
      this._copyValue();
}

File: src/widget/text-print/text-print.js [hunk 2]
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        this.element.addEventListener( 'input', event => {
+            const el = event.target;
+            const widgetText = el.value.replace( /\n/g, '<br>' );
+            widget.innerHTML = widgetText;
+        } );
+    }
+
+    _getClassName() {
```
MartijnR: Consider whether any of this code is still necessary now that we no longer use this for textareas.... :) Try to simplify as much as you can.

src/widget/text-print/text-print.js
```diff
@@ -0,0 +1,48 @@
...
+        this.widget = this.element.parentElement.querySelector( `.${className}` );
+        this._copyValue();
+
+        this.element.addEventListener( 'input', () => {
+            this._copyValue();
+        } );
+    }
+
+    _copyValue() {
+        if (this.element.value !== '') {
```
MartijnR: Oops. Sorry I overlooked this one before! The if() statement keeps a ghost of an old value in the print element if the value in the input is cleared!

See http://localhost:8005/?xform=https://github.com/enketo/enketo-core/files/3852678/print-long-text.xml.txt

1. clear value in first question
2. print 

first two questions on printout show old value but should be empty.

gushil: Ah, I missed that. Thanks!

PR #688, opened 2019-12-16
Replaces enketo-xpathjs

vimemo:
Replaces enketo-xpathjs with openrosa-xpath-evaluator (beta version).


MartijnR:
Woohoooo!! Thanks!! 🎉 

I'll be doing some performance testing and am very curious.

This straightforward replacement will help a lot with figuring out a build system that still can create an IE11 build with the old evaluator.

Since it will break custom non-ODK XPath extensions and the existing build system,  I may try to do one more release before that (in the next few days) to squeeze in a feature a client is waiting for. A 6.0.0 release may be appropriate for this (not sure).

PR #694, opened 2020-02-04
fixed: OC: Print CRF/Casebook PDF long string(s) in text field overlaps

theywa:
#166 
[OC-12196](https://jira.openclinica.com/browse/OC-12196)
example:
[Vital Signs-long-text-single-line.pdf](https://github.com/enketo/enketo-core/files/4152109/Vital.Signs-long-text-single-line.pdf)

### DIFF
--- a/src/sass/core/_print.scss
+++ b/src/sass/core/_print.scss
@@ -455,3 +455,24 @@ h4 {
         }
     }
 }
+
+.or-text-print-initialized {
+    input {
+        &[type=text] {
+            display: none;
+        }
+    }
+
+    .print-input-text {
+        font-weight: normal;
+        font-size: 12px;
+        width: 80%;
+        padding: 6px 12px;
+        border-bottom: 1px solid black;
+    }
+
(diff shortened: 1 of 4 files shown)


theywa:
@MartijnR - yes you're right, I move the media print style on text-print.scss to _print.scss and it fixed the issue. The previous jquery code already removed and will take a note about adding more script for next time(unless it's needed). Thanks for the lesson.
[demo-vitals-grid-v27b.pdf](https://github.com/enketo/enketo-core/files/4184887/demo-vitals-grid-v27b.pdf)

MartijnR:
Thanks @theywa, it looks like the second `@media print` query in grid/_widgets.scss may also cause issues in some edge cases. Might be good to move that in the same PR.

theywa:
sorry, I miss that one, code updated and here the preview
[demo-vitals-grid-v27b.pdf](https://github.com/enketo/enketo-core/files/4191119/demo-vitals-grid-v27b.pdf)

PR #698, opened 2020-02-28
Use absolute import for bootstrap-datepicker

tnagorra:
- Remove background-image for geo-picker

### DIFF
--- a/src/widget/geo/geopicker.scss
+++ b/src/widget/geo/geopicker.scss
@@ -1116,13 +1116,13 @@ $input-button-border: 2px solid #aaaaaa;
 }
 
 .leaflet-control-layers-toggle {
-    background-image: url(images/layers.png);
+    // background-image: url(images/layers.png);
     width: 36px;
     height: 36px;
 }
 
 .leaflet-retina .leaflet-control-layers-toggle {
-    background-image: url(images/layers-2x.png);
+    // background-image: url(images/layers-2x.png);
     background-size: 26px 26px;
 }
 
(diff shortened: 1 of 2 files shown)


src/widget/geo/geopicker.scss
```diff
@@ -1116,13 +1116,13 @@ $input-button-border: 2px solid #aaaaaa;
 }
 
 .leaflet-control-layers-toggle {
-    background-image: url(images/layers.png);
+    // background-image: url(images/layers.png);
```
tnagorra: I could not find the image `layers.png` anywhere in the repository.

MartijnR:
Thank you. Would you perhaps like to do the same for the other image file references in that stylesheet?

tnagorra:
@MartijnR the only other reference to `url()` in that stylesheet I could find is
https://github.com/the-deep/enketo-core/blob/4cb8558ffbc4b44f1258f0280fd68a7d149505eb/src/widget/geo/geopicker.scss#L843-L847

Is it safe to comment this out?

MartijnR:
Oops, sorry. Turns out I didn't review your PR properly and only looked a the commented change above.

tnagorra:
I am using webpack to build the scss and the absolute import is provided by webpack.
I didn't know scss was built using grunt in this project. Will have to look into grunt and grunt-sass to get absolute import working.

MartijnR:
I've merged it, but reverted the import. It would still be welcome, but I don't have time myself to dive into SCSS processors atm and I believe our current one may be faster than the competition, which is quite cool during development. Thanks!

PR #703, opened 2020-03-18
fixed: OC: include textarea on text-print widget selector(#312)

theywa:
[312](https://github.com/OpenClinica/enketo-express-oc/issues/312) [OC-12474](https://jira.openclinica.com/browse/OC-12474)

### DIFF
--- a/src/widget/textarea/textarea.js
+++ b/src/widget/textarea/textarea.js
@@ -23,6 +23,10 @@ class TextareaWidget extends Widget {
             }
         } );
         textareas.forEach( this._resize.bind( this ) );
+        this.element.addEventListener( 'pageflip', event => {
+            const els = event.target.querySelectorAll( 'textarea' );
+            els.forEach( this._resize.bind( this ) );
+        } );
     }
 
     _resize( el ) {


theywa:
Hi @MartijnR,
I see, after rechecking I think the separate pages broke that _resize method(only for loaded values), for example, if the text area is on the second page then it is not rendered yet(CMIIW) so the value of scroll height and client height is 0. I test it using the timeout to check the scroll/client height value. Do you have any idea to fix the issue?

MartijnR:
Ah that explains it. Great you found that. There is a page flip event (`events.PageFlip()`), see pages.js. I think adding a handler for that event in the textarea widget is a good way to fix this.

I think the handler would have to be placed on the page Element ancestor of the `<textarea>` element which could be the question itself or a group ancestor (`[role="page"]`).

theywa:
Got it, your solution works like a charm :smile: 
Added handler on section[role=page] and it fixed the auto-resize on default value and printout.
There is a CSS issue with the text area width when printing but I don't see it on OC(there is some script that added extra inline style on OC to make the width 100%) so I ignore it.
example :
[Vital Signs.pdf](https://github.com/enketo/enketo-core/files/4359624/Vital.Signs.pdf)

PR #706, opened 2020-04-14
fiixed: OC: Issue text area not resize correctly when printing

theywa:
[#312](https://github.com/OpenClinica/enketo-express-oc/issues/312) 

form-model.js and input.js updated when I run `grunt jsbeautifier:fix` should I exclude those files?

### DIFF
--- a/src/sass/core/_print.scss
+++ b/src/sass/core/_print.scss
@@ -457,15 +457,8 @@ h4 {
     }
 }
 
-.or-text-print-initialized {
-    input {
-        &[type=text] {
-            display: none;
-        }
-    }
-
-    .print-only {
-        display: inline-block;
-        page-break-inside: avoid;
+input[type=text], textarea {
+    &.print-only {
+        display: none;
     }
 }
--- a/src/widget/text-print/text-print.js
+++ b/src/widget/text-print/text-print.js
@@ -1,4 +1,5 @@
 import Widget from '../../js/widget';
+import events from '../../js/event';
 
 /**
  * Clone text input fields value into new print-only element.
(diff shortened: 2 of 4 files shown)


theywa:
Code updated

src/sass/core/_print.scss
```diff
@@ -458,10 +458,15 @@ h4 {
...
-        &[type=text] {
+    input[type=text], 
+    textarea {
+        &:not(.default-print) {
             display: none;
         }
+        
+        &.default-print {
+            overflow: hidden;
+        }
```
MartijnR, line 469: I think we can clean up the amount of CSS classes used with this new plan.

I think .or-text-print-initialized no longer serves any purpose (now that we use printify and deprinty events to make temporary modifications during printing only).

I think the name `.default-print` is a little confusing. I'm wondering if we can use just `.print-replacement` (and no longer need `.print-only` and `.print-default`). Then we can target with `.print-only ~ input[type=text], .print-only ~ textarea{}`. Does that sensible to you?

And we could just remove the whole added element and add it only during the printify/deprintify events. Or toggle a .hide/.print-replacement class. Whatever you think is best.

src/widget/text-print/text-print.js
```diff
@@ -25,17 +26,23 @@ class TextPrintWidget extends Widget {
...
 
         this.widget = this.element.parentElement.querySelector( `.${className}` );
-        this._copyValue();
-
-        this.element.addEventListener( 'input', () => {
-            this._copyValue();
-        } );
+        this.element.classList.add( 'default-print' );
+        const comment = this.question.parentElement.querySelector( '[role="comment"]' );
+        $( comment )
```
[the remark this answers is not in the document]
theywa: this is for showing text area/input value when user show print preview without API(e.g. like using ctrl+p) since the text area/input will be hidden on print preview and no value copied to text-print widget yet then without this class, the user will not see any value. This class is not needed previously since there is an input listener to automatically update the text-print widget.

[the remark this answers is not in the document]
MartijnR: I was referring to 'comment'. :)

src/widget/text-print/text-print.js
```diff
@@ -25,17 +26,23 @@ class TextPrintWidget extends Widget {
...
+            .on( 'printify.enketo', this._copyValue.bind( this ) )
+            .on( 'deprintify.enketo', this._resetValue.bind( this ) );
```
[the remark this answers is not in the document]
theywa: Sorry for the late response, code updated : 

- add new custom events
- adding the widget on printify event
- remove the widget on deprintify event
- reduce the css style but keep the style clas as a marker

src/widget/text-print/text-print.js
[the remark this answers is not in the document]
theywa: ah my bad, previously I used comment/dn widget as the listener since the printify event dispatched from that widget.

src/sass/core/_print.scss
theywa: style updated

src/widget/text-print/text-print.js
```diff
@@ -12,32 +13,35 @@ class TextPrintWidget extends Widget {
     static get selector() {
         // The [data-for] exclusion prevents "comment" widgets from being included.
         // It is not quite correct to do this but atm the "for" attribute in XForms is only used for comment widgets.
-        return '.question:not(.or-appearance-autocomplete):not(.or-appearance-url) > input[type=text]:not(.ignore):not([data-for])';
+        return '.question:not(.or-appearance-autocomplete):not(.or-appearance-url) > input[type=text]:not(.ignore):not([data-for]), .question:not(.or-appearance-autocomplete):not(.or-appearance-url) > textarea:not(.ignore):not([data-for])';
     }
 
     _init() {
-        this.question.classList.add( 'or-text-print-initialized' );
+        this.question.classList.add( 'print-input-text' );
```
MartijnR, line 20: I think we don't need any class here. We want to make sure that the real input fields still print if user bypass the print script (which is fine for non-Grid themes) by Cmd-P/Ctrl-P e.g.

src/sass/core/_main.scss
```diff
@@ -419,7 +419,6 @@ legend {
     }
 
     input[type=text],
-    .print-input-text,
```
MartijnR, line 422: It is important to use the same style as for input fields. Not doing this re-creates a bunch of issues that were fixed by  08238fcf242bb0072447945a647535218fa46fdd (and a few later commits)

theywa:
Hi @MartijnR,  sorry for the late update:

- Undoing all CSS changes (check, undo all necessary CSS)
- Don't use a class on the question(check)
- Use the CSS sibling selector I proposed for .print-input-text(I think is check too)

src/sass/core/_main.scss
theywa: got it, undo the CSS changes

src/widget/text-print/text-print.js
theywa: class removed

MartijnR:
Thank you! Sorry, I was wrong about the ~ general sibling-selector!

theywa:
No problem, I also not sure the selector will work so not use that

PR #712, opened 2020-05-04
fixed: regressions with Grid theme print style + refactoring

MartijnR:
Hi @theywa,

Since this is still fresh, I was hoping you could do a sanity check on this PR.

I'm going to put comments inline.

### DIFF
--- a/app.js
+++ b/app.js
@@ -10,6 +10,7 @@ import support from './src/js/support';
 import { Form } from './src/js/form';
 import fileManager from './src/js/file-manager';
 import events from './src/js/event';
+import { fixGrid, styleToAll, styleReset } from './src/js/print';
 var loadErrors;
 var form;
 var formStr;
@@ -69,6 +70,7 @@ function initializeForm() {
     form = new Form( formEl, {
         modelStr: modelStr
     }, {
+        'printRelevantOnly': false,
         'clearIrrelevantImmediately': false
     } );
     // for debugging
@@ -86,3 +88,19 @@ function getURLParameter( name ) {
         ( new RegExp( name + '=' + '(.+?)(&|$)' ).exec( location.search ) || [ null, null ] )[ 1 ]
     );
 }
+
+// to facilitate developing print-specific issues
+function printView( on = true ) {
+    if ( on ) {
+        document.querySelectorAll( '.question' ).forEach( el => el.dispatchEvent( events.PrintifyText() ) );
+        styleToAll();
+        setTimeout( () => {
+            fixGrid( { format: 'letter' } );
+        }, 500 );
+
+    } else {
+        document.querySelectorAll( '.question' ).forEach( el => el.dispatchEvent( events.DePrintifyText() ) );
+        styleReset();
+    }
+}
+window.printView = printView;
--- a/src/sass/core/_print.scss
+++ b/src/sass/core/_print.scss
@@ -31,6 +31,7 @@ body {
     }
 }
 
+.print-hide,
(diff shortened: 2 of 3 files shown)


app.js
```diff
@@ -86,3 +88,19 @@ function getURLParameter( name ) {
         ( new RegExp( name + '=' + '(.+?)(&|$)' ).exec( location.search ) || [ null, null ] )[ 1 ]
     );
 }
+
+// to facilitate developing print-specific issues
```
MartijnR, line 92: app.js is not included in the Enketo Core module (when installing from npm), so I thought it was okay and helpful to add this function for us developer. We can do `printView(false)` and `printView(true)` in the console to switch between screen and print views.

app.js
```diff
@@ -69,6 +70,7 @@ function initializeForm() {
     form = new Form( formEl, {
         modelStr: modelStr
     }, {
+        'printRelevantOnly': false,
```
MartijnR, line 73: We'll probably still have to switch this between false/true, depending on what we are working on it. It determines whether skipped (invisible) questions are shown in the print view or not.

src/sass/core/_print.scss
```diff
@@ -31,6 +31,7 @@ body {
     }
 }
 
+.print-hide,
```
MartijnR, line 34: I prefer this class name (instead of .print-only) as it is more descriptive.

src/widget/text-print/text-print.js
```diff
@@ -22,27 +22,25 @@ class TextPrintWidget extends Widget {
     }
 
     _addWidget() {
-        const className = 'print-text-input';
```
MartijnR, line 25: Note, you had swapped print-input-text with print-text-input. ;) This broke the recent CSS fixes.

src/widget/text-print/text-print.js
```diff
@@ -22,27 +22,25 @@ class TextPrintWidget extends Widget {
...
-        const printElement = document.createElement( 'div' );
-        printElement.classList.add( className, 'widget' );
+        if ( !this.widget ) {
```
MartijnR, line 25: The if() statement just prevents exceptions/duplicate elements in case the printify/deprintify events are not always following each other for any reason.

src/widget/text-print/text-print.js
```diff
@@ -22,27 +22,25 @@ class TextPrintWidget extends Widget {
...
+            const className = 'print-input-text';
+            const printElement = document.createElement( 'div' );
+            printElement.classList.add( className, 'widget' );
 
-        if ( getComputedStyle( this.element ).order !== '' ) {
```
MartijnR, line 29: I believe we no longer need this hack, as the CSS should now take care of this.

src/widget/text-print/text-print.js
```diff
@@ -22,27 +22,25 @@ class TextPrintWidget extends Widget {
...
+            this.widget.innerHTML = this.element.value.replace( /\n/g, '<br>' );
+        }
     }
 
     _removeWidget() {
-        this.element.classList.remove( 'print-only' );
-        this.element.parentElement.removeChild( this.widget );
-    }
-
-    update() {
```
MartijnR, line 44: The update function should remain a no-op here, because the update function is called for several (non-print-related) reasons (see widgets-controller.js) and that would cause problems.

MartijnR:
I'll have to merge this as a dictator, because of the bugs it resolves that are currently in master. However, @theywa, your feedback on this PR is still welcome.

Issue #720, opened 2020-06-05
Editing records within a repeat group


Janemesi:
We are encountering an issue with relevant conditions when editing records on Enketo that are within nested repeat groups. An error message is thrown when one has submitted information for more than 2 iterations on each repeat group. 

For instance, on the sample form shared the error message is being thrown when one submits information for more than 2 buildings and on each building, one submits information for more than 2 rooms. We have tested the same form on KoBo and we were able to replicate the issue. Any recommendations or ideas that you may have regarding resolving this issue will be highly appreciated.

The attached screenshot displays the thrown error message. 
<img width="933" alt="er1" src="https://user-images.githubusercontent.com/36330603/83870196-7e76aa00-a736-11ea-9975-30c3b87f9c05.png">
[SchoolSurveyTestingFOrm2.xlsx](https://github.com/enketo/enketo-express/files/4735683/SchoolSurveyTestingFOrm2.xlsx)

FrankApiyo:
Hey, @MartijnR, I'm interested in working on this. I've been able to set up enketo locally and replicate the issue. I may reach out when I need some guidance.

FrankApiyo:
Although it seems I can't really `git push` to this repo, I already have some changes that I'd like to get your opinion on.

Please let me know if there is anything I need to do in order to be able to push code here. Thank.

MartijnR:
That's great @FrankApiyo! I've invited you to Enketo Collaborators, so I think you can push into a "fix/" branch now. Otherwise, you can also fork and send a PR. Thanks!

PR #723, opened 2020-06-16
Fix errors raised when adding nested repeats

FrankApiyo:
Fixes: https://github.com/enketo/enketo-core/issues/720

# TODO 

- [x]  Add/Fix tests

### DIFF
--- a/src/js/repeat.js
+++ b/src/js/repeat.js
@@ -199,9 +199,10 @@ export default {
     updateDefaultFirstRepeatInstance( repeatInfo ) {
         const repeatPath = repeatInfo.dataset.name;
         if ( !this.form.model.data.instanceStr && !this.templates[ repeatPath ].classList.contains( 'or-appearance-minimal' ) ) {
-            const repeatSeriesIndex = this.getIndex( repeatInfo );
-            const repeatSeriesInModel = this.form.model.getRepeatSeries( repeatPath, repeatSeriesIndex );
-            if ( repeatSeriesInModel.length === 0 ) {
+            const prevSibling = repeatInfo.previousElementSibling;
+            let repeatIndexInSeries = prevSibling && prevSibling.classList.contains( 'or-repeat' ) ?
+                Number( prevSibling.querySelector( '.repeat-number' ).textContent ) : 0;
+            if ( repeatIndexInSeries === 0 ) {
                 this.add( repeatInfo, 1, 'magic' );
             }
 
@@ -282,6 +283,15 @@ export default {
 
         // Determine the index of the repeat series.
         let repeatSeriesIndex = this.getIndex( repeatInfo );
+
+        // get repeatSeriesIndex using parent repeat element
+        // this fixes an nested repeats error
+        if( repeatPath && repeatPath.split( '/' ).length - 1 > 3 && repeatInfo ){
+            let parentRepeatElement = repeatInfo.closest( `[name='${repeatPath.slice( 0, repeatPath.lastIndexOf( '/' ) )}']` );
+            if( parentRepeatElement.parentElement ){
+                repeatSeriesIndex = this.getIndex( parentRepeatElement );
+            }
+        }
         let modelRepeatSeriesLength = this.form.model.getRepeatSeries( repeatPath, repeatSeriesIndex ).length;
         // Determine the index of the repeat inside its series
         const prevSibling = repeatInfo.previousElementSibling;
--- a/src/js/repeat.js
+++ b/src/js/repeat.js
@@ -285,7 +285,7 @@ export default {
         let repeatSeriesIndex = this.getIndex( repeatInfo );
 
         // get repeatSeriesIndex using parent repeat element
-        // this fixes a nested repeats error
+        // this fixes a nested repeats error https://github.com/enketo/enketo-core/issues/720
         if( repeatPath && repeatPath.split( '/' ).length - 1 > 3 && repeatInfo ){
             let parentRepeatElement = repeatInfo.closest( `.or-repeat[name='${repeatPath.slice( 0, repeatPath.lastIndexOf( '/' ) )}']` );
             if( parentRepeatElement && parentRepeatElement.parentElement ){
--- a/src/js/repeat.js
+++ b/src/js/repeat.js
@@ -285,10 +285,10 @@ export default {
         let repeatSeriesIndex = this.getIndex( repeatInfo );
 
         // get repeatSeriesIndex using parent repeat element
-        // this fixes an nested repeats error
+        // this fixes a nested repeats error
         if( repeatPath && repeatPath.split( '/' ).length - 1 > 3 && repeatInfo ){
-            let parentRepeatElement = repeatInfo.closest( `[name='${repeatPath.slice( 0, repeatPath.lastIndexOf( '/' ) )}']` );
-            if( parentRepeatElement.parentElement ){
+            let parentRepeatElement = repeatInfo.closest( `.or-repeat[name='${repeatPath.slice( 0, repeatPath.lastIndexOf( '/' ) )}']` );
+            if( parentRepeatElement && parentRepeatElement.parentElement ){
                 repeatSeriesIndex = this.getIndex( parentRepeatElement );
             }
         }
--- a/src/js/repeat.js
+++ b/src/js/repeat.js
@@ -132,6 +132,17 @@ export default {
         if ( !el || !this.form.repeatsPresent ) {
             return 0;
         }
+
+        // get repeatSeriesIndex using parent repeat element
+        // this fixes a nested repeats error https://github.com/enketo/enketo-core/issues/720
+        const repeatPath = el.dataset.name;
+        if( repeatPath && repeatPath.split( '/' ).length - 1 > 3 && el ){
+            let parentRepeatElement = el.closest( `.or-repeat[name='${repeatPath.slice( 0, repeatPath.lastIndexOf( '/' ) )}']` );
+            if( parentRepeatElement && parentRepeatElement.parentElement ){
+                el = parentRepeatElement;
+            }
+        }
+
         let checkEl = el.parentElement.closest( '.or-repeat' );
         const info = el.classList.contains( 'or-repeat-info' );
         let count = info ? 1 : Number( el.querySelector( '.repeat-number' ).textContent );
@@ -199,10 +210,9 @@ export default {
     updateDefaultFirstRepeatInstance( repeatInfo ) {
         const repeatPath = repeatInfo.dataset.name;
         if ( !this.form.model.data.instanceStr && !this.templates[ repeatPath ].classList.contains( 'or-appearance-minimal' ) ) {
-            const prevSibling = repeatInfo.previousElementSibling;
-            let repeatIndexInSeries = prevSibling && prevSibling.classList.contains( 'or-repeat' ) ?
-                Number( prevSibling.querySelector( '.repeat-number' ).textContent ) : 0;
-            if ( repeatIndexInSeries === 0 ) {
+            const repeatSeriesIndex = this.getIndex( repeatInfo );
+            const repeatSeriesInModel = this.form.model.getRepeatSeries( repeatPath, repeatSeriesIndex );
+            if ( repeatSeriesInModel.length === 0 ) {
                 this.add( repeatInfo, 1, 'magic' );
             }
 
@@ -284,14 +294,6 @@ export default {
         // Determine the index of the repeat series.
         let repeatSeriesIndex = this.getIndex( repeatInfo );
 
-        // get repeatSeriesIndex using parent repeat element
-        // this fixes a nested repeats error https://github.com/enketo/enketo-core/issues/720
-        if( repeatPath && repeatPath.split( '/' ).length - 1 > 3 && repeatInfo ){
-            let parentRepeatElement = repeatInfo.closest( `.or-repeat[name='${repeatPath.slice( 0, repeatPath.lastIndexOf( '/' ) )}']` );
-            if( parentRepeatElement && parentRepeatElement.parentElement ){
-                repeatSeriesIndex = this.getIndex( parentRepeatElement );
-            }
-        }
         let modelRepeatSeriesLength = this.form.model.getRepeatSeries( repeatPath, repeatSeriesIndex ).length;
         // Determine the index of the repeat inside its series
         const prevSibling = repeatInfo.previousElementSibling;
--- a/src/js/repeat.js
+++ b/src/js/repeat.js
@@ -293,7 +293,6 @@ export default {
 
         // Determine the index of the repeat series.
         let repeatSeriesIndex = this.getIndex( repeatInfo );
-
         let modelRepeatSeriesLength = this.form.model.getRepeatSeries( repeatPath, repeatSeriesIndex ).length;
         // Determine the index of the repeat inside its series
         const prevSibling = repeatInfo.previousElementSibling;
--- a/src/js/repeat.js
+++ b/src/js/repeat.js
@@ -133,27 +133,17 @@ export default {
             return 0;
         }
 
-        // get repeatSeriesIndex using parent repeat element
-        // this fixes a nested repeats error https://github.com/enketo/enketo-core/issues/720
-        const repeatPath = el.dataset.name;
-        if( repeatPath && repeatPath.split( '/' ).length - 1 > 3 && el ){
-            let parentRepeatElement = el.closest( `.or-repeat[name='${repeatPath.slice( 0, repeatPath.lastIndexOf( '/' ) )}']` );
-            if( parentRepeatElement && parentRepeatElement.parentElement ){
-                el = parentRepeatElement;
-            }
-        }
-
         let checkEl = el.parentElement.closest( '.or-repeat' );
         const info = el.classList.contains( 'or-repeat-info' );
         let count = info ? 1 : Number( el.querySelector( '.repeat-number' ).textContent );
-        let name;
+        const name = el.dataset.name || el.getAttribute( 'name' );
+
         while ( checkEl ) {
             while ( checkEl.previousElementSibling && checkEl.previousElementSibling.matches( '.or-repeat' ) ) {
                 checkEl = checkEl.previousElementSibling;
                 if ( info ) {
-                    count++;
+                    count += checkEl.querySelectorAll( `.or-repeat-info[data-name="${name}"]` ).length;
                 } else {
-                    name = name || el.getAttribute( 'name' );
                     count += checkEl.querySelectorAll( `.or-repeat[name="${name}"]` ).length;
                 }
             }
--- a/test/forms/nested-repeats.xml
+++ b/test/forms/nested-repeats.xml
@@ -1,0 +1,156 @@
+<?xml version="1.0"?>
+<h:html xmlns="http://www.w3.org/2002/xforms"
+    xmlns:ev="http://www.w3.org/2001/xml-events"
+    xmlns:h="http://www.w3.org/1999/xhtml"
+    xmlns:jr="http://openrosa.org/javarosa"
+    xmlns:odk="http://www.opendatakit.org/xforms"
+    xmlns:orx="http://openrosa.org/xforms"
+    xmlns:xsd="http://www.w3.org/2001/XMLSchema">
+    <h:head>
+        <h:title>School Survey Testing Form</h:title>
+        <model odk:xforms-version="1.0.0">
+            <itext>
+                <translation default="true()" lang="English">
+                    <text id="/data/repeat1:label">
+                        <value>Repeat 1 - Building</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11:label">
+                        <value>Repeat 11 - Corridor</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111:label">
+                        <value>Repeat 111 - Rooms</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/room_open_stat:label">
+                        <value>Is the room open or able to open it?</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/room_open_stat/yes:label">
+                        <value>Yes</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/room_open_stat/no:label">
+                        <value>No</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/school_roomtype:label">
+                        <value>121. Room Type</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/school_roomtype:hint">
+                        <value>Please choose from the list below</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/school_roomtype/roomtype_list1:label">
+                        <value>Library</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/school_roomtype/roomtype_list2:label">
+                        <value>Teacher room</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/school_roomtype/roomtype_list3:label">
+                        <value>Adminstrative Office</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/school_roomtype/roomtype_list4:label">
+                        <value>Specialty room</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/school_roomtype/roomtype_list5:label">
+                        <value>Classroom</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/school_roomtype/roomtype_list6:label">
+                        <value>Computer lab</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/school_roomtype/roomtype_list7:label">
+                        <value>Science Lab</value>
+                    </text>
+                    <text id="/data/repeat1/repeat11/repeat111/school_roomtype/roomtype_list8:label">
+                        <value>Other</value>
+                    </text>
+                </translation>
+            </itext>
+            <instance>
+                <data id="test_school_form4">
+                    <repeat1 jr:template="">
+                        <repeat11 jr:template="">
+                            <repeat111 jr:template="">
+                                <room_open_stat/>
+                                <school_roomtype/>
+                            </repeat111>
+                        </repeat11>
+                    </repeat1>
+                    <repeat1>
+                        <repeat11>
+                            <repeat111>
+                                <room_open_stat/>
+                                <school_roomtype/>
+                            </repeat111>
+                        </repeat11>
+                    </repeat1>
+                    <meta>
+                        <instanceID/>
+                    </meta>
+                </data>
+            </instance>
+            <bind nodeset="/data/repeat1/repeat11/repeat111/room_open_stat" type="string"/>
+            <bind nodeset="/data/repeat1/repeat11/repeat111/school_roomtype" relevant="selected( ../room_open_stat ,'yes')" type="string"/>
+            <bind jr:preload="uid" nodeset="/data/meta/instanceID" readonly="true()" type="string"/>
+        </model>
+    </h:head>
+    <h:body>
+        <group ref="/data/repeat1">
+            <label ref="jr:itext('/data/repeat1:label')"/>
+            <repeat nodeset="/data/repeat1">
+                <group ref="/data/repeat1/repeat11">
+                    <label ref="jr:itext('/data/repeat1/repeat11:label')"/>
+                    <repeat nodeset="/data/repeat1/repeat11">
+                        <group ref="/data/repeat1/repeat11/repeat111">
+                            <label ref="jr:itext('/data/repeat1/repeat11/repeat111:label')"/>
+                            <repeat nodeset="/data/repeat1/repeat11/repeat111">
+                                <select1 ref="/data/repeat1/repeat11/repeat111/room_open_stat">
+                                    <label ref="jr:itext('/data/repeat1/repeat11/repeat111/room_open_stat:label')"/>
+                                    <item>
+                                        <label ref="jr:itext('/data/repeat1/repeat11/repeat111/room_open_stat/yes:label')"/>
+                                        <value>yes</value>
+                                    </item>
+                                    <item>
+                                        <label ref="jr:itext('/data/repeat1/repeat11/repeat111/room_open_stat/no:label')"/>
+                                        <value>no</value>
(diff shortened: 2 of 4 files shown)


MartijnR:
Thank you! Did you find out why the selector does not find any nodes there and in what part of the form initialization? 

It's a little suspicious for this happen upon loading an existing record and I want to make sure we're not obscuring a deeper issue (e.g. with how nested repeats are added).

FrankApiyo:
It seems the selector only misses/does not find nodes when looking for specific nodes. These are nodes that are displayed based on a relevant condition. when I add these nodes manually by editing the XML, this error seems to disappear... I think that another way to solve this would be to ensure that these nodes are always added when adding repeats.

The nodes I added belong to a field called `school_roomtype` with a binding that looks as follows:

`<bind nodeset="/data/group1/repeat1/repeat11/repeat111/school_roomtype" relevant="selected( ../room_open_stat ,'yes')" type="string"/>`

`school_roomtype` nodes added manually

```
<instance>
                <data id="test_school_form">
                    <start_time/>
                    <end_time/>
                    <device_id/>
                    <group1>
                        <repeat1>
                            <position_repeat1>1</position_repeat1>
                            <numb_buid99>1</numb_buid99>
                            <building>bulding_type_list7</building>
                            <repeat11>
                                <position_repeat11>1</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                    <school_roomtype/>
                                </repeat111>
                                <repeat111>
                                    <position_repeat111>2</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                    <school_roomtype/>
                                </repeat111>
                                <total_number_room>1</total_number_room>
                            </repeat11>
                            <repeat11>
                                <position_repeat11>2</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                    <school_roomtype/>
                                </repeat111>
                                <total_number_room>6</total_number_room>
                            </repeat11>
                            <total_number_cord>1</total_number_cord>
                        </repeat1>
                        <repeat1>
                            <position_repeat1>2</position_repeat1>
                            <numb_buid99>1</numb_buid99>
                            <building>bulding_type_list4</building>
                            <repeat11>
                                <position_repeat11>1</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                    <school_roomtype/>
                                </repeat111>
                                <total_number_room>2</total_number_room>
                            </repeat11>
                            <total_number_cord>2</total_number_cord>
                        </repeat1>
                        <repeat1>
                            <position_repeat1>3</position_repeat1>
                            <numb_buid99>1</numb_buid99>
                            <building>bulding_type_list7</building>
                            <repeat11>
                                <position_repeat11>1</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                    <school_roomtype/>
                                </repeat111>
                                <total_number_room>3</total_number_room>
                            </repeat11>
                            <total_number_cord>3</total_number_cord>
                        </repeat1>
                        <repeat1>
                            <position_repeat1>4</position_repeat1>
                            <numb_buid99>1</numb_buid99>
                            <building>bulding_type_list4</building>
                            <repeat11>
                                <position_repeat11>1</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                    <school_roomtype/>
                                </repeat111>
                                <total_number_room>4</total_number_room>
                            </repeat11>
                            <total_number_cord>4</total_number_cord>
                        </repeat1>
                    </group1>
                    <total_numb_b/>
                    <meta>
                        <instanceID/>
                    </meta>
                </data>
```

`school_roomtype` nodes not added 

```
 <data id="test_school_form">
                    <start_time/>
                    <end_time/>
                    <device_id/>
                    <group1>
                        <repeat1>
                            <position_repeat1>1</position_repeat1>
                            <numb_buid99>1</numb_buid99>
                            <building>bulding_type_list7</building>
                            <repeat11>
                                <position_repeat11>1</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                </repeat111>
                                <repeat111>
                                    <position_repeat111>2</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                </repeat111>
                                <total_number_room>1</total_number_room>
                            </repeat11>
                            <repeat11>
                                <position_repeat11>2</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                </repeat111>
                                <total_number_room>6</total_number_room>
                            </repeat11>
                            <total_number_cord>1</total_number_cord>
                        </repeat1>
                        <repeat1>
                            <position_repeat1>2</position_repeat1>
                            <numb_buid99>1</numb_buid99>
                            <building>bulding_type_list4</building>
                            <repeat11>
                                <position_repeat11>1</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                </repeat111>
                                <total_number_room>2</total_number_room>
                            </repeat11>
                            <total_number_cord>2</total_number_cord>
                        </repeat1>
                        <repeat1>
                            <position_repeat1>3</position_repeat1>
                            <numb_buid99>1</numb_buid99>
                            <building>bulding_type_list7</building>
                            <repeat11>
                                <position_repeat11>1</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                </repeat111>
                                <total_number_room>3</total_number_room>
                            </repeat11>
                            <total_number_cord>3</total_number_cord>
                        </repeat1>
                        <repeat1>
                            <position_repeat1>4</position_repeat1>
                            <numb_buid99>1</numb_buid99>
                            <building>bulding_type_list4</building>
                            <repeat11>
                                <position_repeat11>1</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                </repeat111>
                                <total_number_room>4</total_number_room>
                            </repeat11>
                            <total_number_cord>4</total_number_cord>
                        </repeat1>
                    </group1>
                    <total_numb_b/>
                    <meta>
                        <instanceID/>
                    </meta>
                </data>
```


I'm not sure if this makes sense... Please let me know if it does.

I'm currently trying to find out how repeats are added when you click on the add repeat button. After which I'll find out when the nodes that are only shown based on relevant conditions are added to new repeats and when they are not added.

MartijnR:
Aha. Thanks. The code for addRepeat is in repeat.js (and it eventually calls [addRepeat](https://enketo.github.io/enketo-core/FormModel.html#addRepeat) in form-model.js which adds them to the model.

After your explanation, I actually suspect this issue might be with merging a record into the default model. This also happens in [form-model.js](https://enketo.github.io/enketo-core/FormModel.html#mergeXml). 

I wonder if it would be most efficient to first add your real-life case as a test here: https://github.com/enketo/enketo-core/blob/master/test/spec/formmodel.spec.js#L1091 to see if the merged model is correct.

If it fails, you could simplify the failing test, and then probably the fix would be in the mergeXml function. If it doesn't fail, look elsewhere.

MartijnR:
P.S. Those tests are pre-ES6 and hard to read. Definitely don't feel obliged to not use 

```xml
`<much>
     <more>
          <readable>
                 <syntax/>
          </readable>
     </more>
</much>`
```
using template literals...

FrankApiyo:
In order to create a submission that when editing will raise this error, the following error occurs while filling in the form. I believe it is the reason why repeat group XML is created incorrectly:

```
enketo-bundle.js:52368 Uncaught TypeError: Cannot read property 'querySelector' of undefined
    at enketo-bundle.js:52368
    at Array.forEach (<anonymous>)
    at Object.update (enketo-bundle.js:52343)
    at enketo-bundle.js:53555
    at Array.forEach (<anonymous>)
    at HTMLFormElement.<anonymous> (enketo-bundle.js:53554)
    at Object.add (enketo-bundle.js:25248)
    at HTMLButtonElement.<anonymous> (enketo-bundle.js:25013)
    at HTMLFormElement.dispatch (enketo-bundle.js:6376)
    at HTMLFormElement.elemData.handle (enketo-bundle.js:6180)
```

This error occurs here: https://github.com/enketo/enketo-core/blob/master/src/js/calculate.js#L59

The values for `repeatPath`, `repeatIndex` and `updated` are as follows when the error occures:

```
repeatPath:  /data/group1/repeat1/repeat11/repeat111
repeatIndex:  8
updated:  {repeatIndex: 8, repeatPath: "/data/group1/repeat1/repeat11/repeat111", trigger: "user", cloned: true}
```

I think this is because repeatIndex has an incorrect value, I'll investigate why this value is incorrect and what it should be.

FrankApiyo:
I now do not think that this is caused by enketo merging records into the default model incorrectly. This is because I have made a submission with ODK and noted that this submission made by ODK collect also has a similar issue.

MartijnR:
The record with nested repeats may have been submitted correctly by either Enketo or Collect. 

Collect may not submit empty non-relevant nodes. Enketo does. Both are correct.

However, when loading a record without those absent non-relevant (repeat) nodes into Enketo, the merge function is supposed to add those missing missing nodes. I think you'll find that's happening correctly with top-level repeats, but not with nested repeats. If so, that's where the fix should be.

If at any point Enketo has a model state (`form.model.xml`) with missing nodes (inside a repeat), that's the bug, as that is supposed to be impossible to occur.

FrankApiyo:
Hey @MartijnR, I'm unsure about this but I think the merge function you are speaking about here is the one called [here](https://github.com/enketo/enketo-core/blob/4bc20e01fb96300ce0128072171bb842cbd5b74d/src/js/form-model.js#L185), `data.instanceStr`, I could not find any place the variable checked before it's called is set.

MartijnR:
Hi @FrankApiyo, It's an optional property of the [FormDataObj](https://enketo.github.io/enketo-core/global.html#FormDataObj), that is used when instantiating a [Form](https://enketo.github.io/enketo-core/Form.html).

So in enketo-core you'd add it [here](https://github.com/enketo/enketo-core/blob/master/app.js#L72) just like modelStr. The value is the (string of) the XML record that is stored by Ona and passed to Enketo whenever the user clicks the edit button.

FrankApiyo:
Your explanation has led me to think the bug is in [this](https://github.com/enketo/enketo-core/blob/master/src/js/form-model.js#L328) function someplace.

MartijnR:
Right. It's likely here: https://github.com/enketo/enketo-core/blob/master/src/js/form-model.js#L379-L414

MartijnR:
LOL, this comment is particular suspicious... https://github.com/enketo/enketo-core/blob/master/src/js/form-model.js#L404

FrankApiyo:
That made me laugh when I saw it too. and this one https://github.com/enketo/enketo-core/blob/master/src/js/form-model.js#L351

MartijnR:
Haha, yes, there was clearly some frustration! :D

FrankApiyo:
Hey @MartijnR, so I have found a few things I'm uncertain about in the `form-model.js` file.

1.  I've added a file called `roomrepeats.xml` which I'm using to test. When I open this file with enketo-core, it seems [this](https://github.com/enketo/enketo-core/blob/master/src/js/form-model.js#L185) model instance variable is always undefined, ergo  `mergeXml` on the next line is never called. (I think there's a probability that the form is initialized differently when running enketo express so this may not be a problem. However, this is unclear to me since it seems essential that we call mergeXml.)
I have changed these lines in this branch in order to get to the `mergeXml` call as you can see [here](https://github.com/enketo/enketo-core/pull/723/files#diff-270701b55df55a32341fa63d95b2d5e3L185-R186)

2. Later on, I noticed that the value of [`repeatPaths`](https://github.com/enketo/enketo-core/blob/master/src/js/form-model.js#L401) here is always undefined too (at least for the form that I have added to the examples). This is clearly a problem. I'm not certain what the regular expression is matching for exactly; if possible you can elaborate on this.

I plan to stay up late today so I can have a more realtime-ish discussion with you on this if it'll be possible

Thanks

MartijnR:
Hi Frank, I link above to this: https://github.com/enketo/enketo-core/blob/master/app.js#L72 (just add the instanceStr property as I described above).

FrankApiyo:
Thanks, I really appreciate you replying.

I'm uncertain what `this.templates` [here](https://github.com/enketo/enketo-core/blob/master/src/js/form-model.js#L47) was/is meant to be/contain. It is never really explicitly assigned anything other than `{}`.

MartijnR:
Hi Frank, it's populated here: https://github.com/enketo/enketo-core/blob/master/src/js/form-model.js#L758. 

The templates are clones of a default repeat in the model. So when a repeat is added (and also for the first repeat, depending on the pyxform version used), it is created from the template. For pyxform-created forms, you'll find `r:template=""` in the primary instance which indicates the structure of the repeat.

Hope that clarifies it. Otherwise, let me know!

FrankApiyo:
Do you think this is an ok approach to fix this issue? https://github.com/enketo/enketo-core/pull/723/files#diff-270701b55df55a32341fa63d95b2d5e3R362-R374

The obstacle to implementing this for me right now is Identifying which nodes are `relevant` nodes. I know that this can be easily done by checking the bindings in the model. Although the form-model's `.xml` field seems to be missing the bindings. Do youknow where I can find them or if they are available in the form model?

app.js
```diff
@@ -69,7 +69,8 @@ document.querySelector( '#validate-form' ).addEventListener( 'click', () => {
 function initializeForm() {
     const formEl = document.querySelector( 'form.or' );
     form = new Form( formEl, {
-        modelStr: modelStr
+        modelStr: modelStr,
+        instanceStr: new DOMParser().parseFromString( modelStr, 'text/xml' ).querySelector( 'model > instance' ).innerHTML
```
MartijnR, line 73: I don't understand what your purpose is here... The bug is related to the record, that you're editing. So you have to actually use the complete record that is creating the problem.

src/js/form-model.js
```diff
@@ -359,6 +359,19 @@ FormModel.prototype.mergeXml = function( recordStr ) {
...
+    //get the parent element of the relevant node
+    //get node that comes before this relevant node in template elements
+    // get all the parent elements of the nodes in the record
+    // var relevantParents = record.querySelectorAll("repeat111");
+    //check if one of it's children is the relevant node, and if not then
+    // insert the relevant node after the node above for each parent element
+
+    // repeat for all relevant node
+    // - we need to first find all relevant nodes
+
```
MartijnR, line 374: No, this would not be the right approach. I suspect the issue is with merging the record into the model, so that's where the fix would be.

app.js
FrankApiyo: If I do not pass this value here to `instanceStr` then it stays undefined and then `mergeXml` is never called [here](https://github.com/enketo/enketo-core/blob/master/src/js/form-model.js#L186). Unless you mean I'm assigning the wrong value to it, it seems important to me that `instanceStr` is assigned a value. Please let me know what you think about this. Thanks

MartijnR: It should be the **actual record** that is causing the problem. How are you otherwise going to reproduce the problem? 

So like: 

```js
instanceStr: `<data id="test_school_form">
                    <start_time/>
                    <end_time/>
                    <device_id/>
                    <group1>
                        <repeat1>
                            <position_repeat1>1</position_repeat1>
                            <numb_buid99>1</numb_buid99>
                            <building>bulding_type_list7</building>
                            <repeat11>
                                <position_repeat11>1</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                </repeat111>
                                <repeat111>
                                    <position_repeat111>2</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                </repeat111>
                                <total_number_room>1</total_number_room>
                            </repeat11>
                            <repeat11>
                                <position_repeat11>2</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                </repeat111>
                                <total_number_room>6</total_number_room>
                            </repeat11>
                            <total_number_cord>1</total_number_cord>
                        </repeat1>
                        <repeat1>
                            <position_repeat1>2</position_repeat1>
                            <numb_buid99>1</numb_buid99>
                            <building>bulding_type_list4</building>
                            <repeat11>
                                <position_repeat11>1</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                </repeat111>
                                <total_number_room>2</total_number_room>
                            </repeat11>
                            <total_number_cord>2</total_number_cord>
                        </repeat1>
                        <repeat1>
                            <position_repeat1>3</position_repeat1>
                            <numb_buid99>1</numb_buid99>
                            <building>bulding_type_list7</building>
                            <repeat11>
                                <position_repeat11>1</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                </repeat111>
                                <total_number_room>3</total_number_room>
                            </repeat11>
                            <total_number_cord>3</total_number_cord>
                        </repeat1>
                        <repeat1>
                            <position_repeat1>4</position_repeat1>
                            <numb_buid99>1</numb_buid99>
                            <building>bulding_type_list4</building>
                            <repeat11>
                                <position_repeat11>1</position_repeat11>
                                <total_numb_corridors>1</total_numb_corridors>
                                <repeat111>
                                    <position_repeat111>1</position_repeat111>
                                    <total_numb_rooms>1</total_numb_rooms>
                                    <room_open_stat>no</room_open_stat>
                                </repeat111>
                                <total_number_room>4</total_number_room>
                            </repeat11>
                            <total_number_cord>4</total_number_cord>
                        </repeat1>
                    </group1>
                    <total_numb_b/>
                    <meta>
                        <instanceID/>
                    </meta>
                </data>`,

```

MartijnR: Note that app.js is just a minimal template for developers to be able to run enketo-core (which is a library, not an app) locally without enketo-express.

So this edit is just for during development allowing you to reproduce and fix the bug.

MartijnR:
This is what I recommend as the next step to resolve this (copied from above):

> I wonder if it would be most efficient to first add your real-life case as a test here: https://github.com/enketo/enketo-core/blob/master/test/spec/formmodel.spec.js#L1091 to see if the merged model is correct.

> If it fails, you could simplify the failing test, and then probably the fix would be in the mergeXml function. If it doesn't fail, look elsewhere.

Of course, you could also just start with a minimal test case right away by merging `<data><rep1><rep2><rep3><a/></rep3></rep>2</rep1></data>` or something.

Please add the test to the PR.

test/spec/formmodel.spec.js
```diff
@@ -1166,6 +1166,11 @@ describe( 'merging an instance into the model', () => {
             // empty group in record
             [ '<data><age>7</age><details/></data>', '<model><instance><data><age/><details><name>Baby ...</name></details></data></instance></model>',
                 '<model><instance><data><age>7</age><details><name/></details></data></instance></model>'
+            ],
+            [
+                "<data><rep1><rep2><rep3><a/></rep3></rep>2</rep1></data>",
+                '<model><instance><data><rep1 jr:template=""><rep2 jr:template=""><rep3 jr:template=""></a></rep3></rep2></rep1></data></model></model>',
```
MartijnR, line 1172: double `</model>`, and no `</instance>`

test/spec/formmodel.spec.js
```diff
@@ -1166,6 +1166,11 @@ describe( 'merging an instance into the model', () => {
...
+                "<model><instance><data><rep1><rep2><rep3><a/></rep3></rep>2</rep1></data></model></model>"
```
MartijnR, line 1173: same

MartijnR: so this test fails (which is what you wanted) but not for the right reason ;)

MartijnR:
That's unfortunate that the test is not failing...

FrankApiyo:
I made a change here that seems to remove the error thrown when adding new repeats https://github.com/enketo/enketo-core/pull/723/commits/a82271dcdb99ca01e6152e32b8f8de8e7035de84.  This removes some errors that enketo-core might throw when you try to add new instances of the middle repeat group.

Let me know if you have any thoughts on this

MartijnR:
@FrankApiyo, a whole bunch of existing tests are failing now (run `npm test`). This makes sense to me because the edit to getIndex breaks it. 

I think, the next step should (still) be to change the test you added to reproduce the issue.

FrankApiyo:
it has not become clear to me that this issue probably being caused my missing elements in the model XML. In the XML snippet below, the last **repeat11** does not contain all the children it should(expected to) contain:
``` 
<data xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms" id="test_school_form4">
                    <!--repeat://data/repeat1-->
                    <repeat1>
                        <!--repeat://data/repeat1/repeat11--><repeat11>
                            <!--repeat://data/repeat1/repeat11/repeat111--><repeat111>
                                <room_open_stat/>
                                <school_roomtype/>
                            </repeat111>
                        </repeat11><repeat11>
                            <!--repeat://data/repeat1/repeat11/repeat111--><repeat111>
                                <room_open_stat/>
                                <school_roomtype/>
                            </repeat111>
                        </repeat11><repeat11>
                            <!--repeat://data/repeat1/repeat11/repeat111--><repeat111>
                                <room_open_stat/>
                                <school_roomtype/>
                            </repeat111>
                        </repeat11>
                    </repeat1><repeat1>
                        <!--repeat://data/repeat1/repeat11--><repeat11>
                            <!--repeat://data/repeat1/repeat11/repeat111-->
                        </repeat11>
                    </repeat1>
                    <meta>
                        <instanceID>uuid:7e6f6c98-a53a-43d4-b03e-8741e65c1b18</instanceID>
                    </meta>
```

I have tried to look around but could not find the code that is responsible for adding to this instance, the xml for repeats when you click the `+` button, because I think that code could be where the issue is

FrankApiyo:
@MartijnR this currently fixes the issue locally for me. let me know what your thoughts are

MartijnR:
@FrankApiyo, I appreciate your work, but note this PR will not be reviewed until:

* there is a failing test that is fixed by your work
* there are no failing existing tests (there are a bunch, see Travis)

FrankApiyo:
> @FrankApiyo, I appreciate your work, but note this PR will not be reviewed until:
> 
> * there is a failing test that is fixed by your work
> * there are no failing existing tests (there are a bunch, see Travis)

Will work on it

FrankApiyo:
@MartijnR  this is ready for a review

MartijnR:
Thanks! I will definitely look at it with interest. Note, that I sometimes only work a couple of hours a week on Enketo, so it could sometimes take a while.

src/js/repeat.js
```diff
@@ -199,9 +199,10 @@ export default {
     updateDefaultFirstRepeatInstance( repeatInfo ) {
         const repeatPath = repeatInfo.dataset.name;
         if ( !this.form.model.data.instanceStr && !this.templates[ repeatPath ].classList.contains( 'or-appearance-minimal' ) ) {
-            const repeatSeriesIndex = this.getIndex( repeatInfo );
-            const repeatSeriesInModel = this.form.model.getRepeatSeries( repeatPath, repeatSeriesIndex );
-            if ( repeatSeriesInModel.length === 0 ) {
+            const prevSibling = repeatInfo.previousElementSibling;
+            let repeatIndexInSeries = prevSibling && prevSibling.classList.contains( 'or-repeat' ) ?
+                Number( prevSibling.querySelector( '.repeat-number' ).textContent ) : 0;
+            if ( repeatIndexInSeries === 0 ) {
```
MartijnR, line 205: I'm a little lost with what is happening here. Could you explain please? Did you consider fixing this in model.getRepeatSeries instead? Doesn't that function have a bug that you discovered here? If so, would it be best to fix that function instead of working around it, or am I misunderstanding?

src/js/repeat.js
```diff
@@ -282,6 +283,15 @@ export default {
 
         // Determine the index of the repeat series.
         let repeatSeriesIndex = this.getIndex( repeatInfo );
+
+        // get repeatSeriesIndex using parent repeat element
+        // this fixes a nested repeats error https://github.com/enketo/enketo-core/issues/720
+        if( repeatPath && repeatPath.split( '/' ).length - 1 > 3 && repeatInfo ){
```
MartijnR, line 289: So, if I understand correctly the bug starts with triple-nested repeats (and continues in quadruple-nested repeats etc). Is that right?

Isn't the actual bug in `this.getIndex`?  If so, would it be best to fix that function instead of working around it for more-than-double-nested repeats, or am I misunderstanding?

test/forms/nested-repeats.xml
```diff
@@ -0,0 +1,156 @@
+<?xml version="1.0"?>
```
MartijnR, line 1: This is okay, but I would normally use a much much smaller form, so keep these tests easier to understand. E.g. no itext, only 1 or 2 select options (or just text inputs).

test/spec/repeat.spec.js
```diff
@@ -70,6 +70,26 @@ describe( 'repeat functionality', () => {
...
+
+        // https://github.com/enketo/enketo-core/issues/720
+        it( 'nested repeats are added correctly', () => {
+            const form = loadForm( 'nested-repeats.xml');
+            form.init();
+
+            // add repeats by clicking the add buttons
+            form.view.$.find( '.or-repeat-info[data-name="/data/repeat1/repeat11/repeat111"] .add-repeat-btn' ).click();
+            form.view.$.find( '.or-repeat-info[data-name="/data/repeat1/repeat11"] .add-repeat-btn' ).click();
+            form.view.$.find( '.or-repeat-info[data-name="/data/repeat1"] .add-repeat-btn' ).click();
```
MartijnR, line 82: Gee,  I played around with this, and this bug is pretty hard to reproduce indeed!

I would use form.view.html.querySelector here, since you want to click only the first button. You do the same but it is less safe.

test/spec/repeat.spec.js
```diff
@@ -70,6 +70,26 @@ describe( 'repeat functionality', () => {
...
+
+            expect(
+                form.getDataStr().replace( />\s+</g, '><' )
+            ).toContain( '<repeat1><repeat11><repeat111><room_open_stat/><school_roomtype/></repeat111><repeat111><room_open_stat/><school_roomtype/></repeat111></repeat11><repeat11><repeat111><room_open_stat/><school_roomtype/></repeat111></repeat11></repeat1><repeat1><repeat11><repeat111><room_open_stat/><school_roomtype/></repeat111></repeat11></repeat1>' );
```
MartijnR, line 86: So in the end after your investigation found that there is no issue when loading a record, right? The issue was with the creation of the **original record** instead, right?

If not, please add a test that reproduces the issue when loading a record (using the same form).

MartijnR (changes requested):

MartijnR:
@lognaturel, @eyelidlessness this is a particularly tough one in my opinion since it deals with nested repeats (and who knows may also fix other nested repeat issues we have). I just wanted to let you know in case, you'd be interested in looking at it, but feel free to ignore as well!

test/spec/repeat.spec.js
FrankApiyo: yes, it seems like this only occurs with creation of a new record or while modifying an existing record by adding nested repeats

src/js/repeat.js
FrankApiyo: makes sense; going to move this code into `this.getIndex`

src/js/repeat.js
FrankApiyo: seems these changes were to get around the error in the `getIndex` function. not needed anymore

test/forms/nested-repeats.xml
FrankApiyo: tried to simplify the form

MartijnR (changes requested):

MartijnR:
@FrankApiyo, have a look at the fix inside the loop that I committed. It turns out the bug was in the performance optimization.

src/js/repeat.js
```diff
@@ -132,17 +132,18 @@ export default {
...
-        let name;
+        const name = el.dataset.name || el.getAttribute( 'name' );
+
         while ( checkEl ) {
             while ( checkEl.previousElementSibling && checkEl.previousElementSibling.matches( '.or-repeat' ) ) {
                 checkEl = checkEl.previousElementSibling;
                 if ( info ) {
-                    count++;
+                    count += checkEl.querySelectorAll( `.or-repeat-info[data-name="${name}"]` ).length;
                 } else {
```
FrankApiyo, line 146: looks good!

MartijnR:
merged!

PR #737, opened 2020-08-16
Add Vertical VAS appearance to show labels every 10 units

gushil:
close https://github.com/OpenClinica/enketo-express-oc/issues/359

### DIFF
--- a/src/widget/analog-scale/analog-scalepicker.js
+++ b/src/widget/analog-scale/analog-scalepicker.js
@@ -133,12 +133,14 @@ class AnalogScaleWidget extends RangeWidget {
     get props() {
         const props = this._props;
         props.touch = support.touch;
+        props.showScale = props.appearances.includes( 'show-scale' );
         props.min = isNumber( this.element.getAttribute( 'min' ) ) ? this.element.getAttribute( 'min' ) : 0;
         props.max = isNumber( this.element.getAttribute( 'max' ) ) ? this.element.getAttribute( 'max' ) : 100;
-        props.step = isNumber( this.element.getAttribute( 'step' ) ) ? this.element.getAttribute( 'step' ) : 1; //( props.type === 'decimal' ? 0.1 : 1 );
+        props.step = isNumber( this.element.getAttribute( 'step' ) ) ? this.element.getAttribute( 'step' ) : ( props.showScale ? 10 : 1 ); //( props.type === 'decimal' ? 0.1 : 1 );
         props.vertical = !props.appearances.includes( 'horizontal' );
         props.ticks = !props.appearances.includes( 'no-ticks' );
         props.maxTicks = 10;
+        props.distress = props.showScale;
 
         return props;
     }
--- a/src/widget/analog-scale/analog-scalepicker.js
+++ b/src/widget/analog-scale/analog-scalepicker.js
@@ -140,7 +140,6 @@ class AnalogScaleWidget extends RangeWidget {
         props.vertical = !props.appearances.includes( 'horizontal' );
         props.ticks = !props.appearances.includes( 'no-ticks' );
         props.maxTicks = 10;
-        props.distress = props.showScale;
 
         return props;
     }
--- a/src/widget/analog-scale/analog-scalepicker.js
+++ b/src/widget/analog-scale/analog-scalepicker.js
@@ -134,9 +134,12 @@ class AnalogScaleWidget extends RangeWidget {
         const props = this._props;
         props.touch = support.touch;
         props.showScale = props.appearances.includes( 'show-scale' );
-        props.min = isNumber( this.element.getAttribute( 'min' ) ) ? this.element.getAttribute( 'min' ) : 0;
-        props.max = isNumber( this.element.getAttribute( 'max' ) ) ? this.element.getAttribute( 'max' ) : 100;
-        props.step = isNumber( this.element.getAttribute( 'step' ) ) ? this.element.getAttribute( 'step' ) : ( props.showScale ? 10 : 1 ); //( props.type === 'decimal' ? 0.1 : 1 );
+        const min = isNumber( this.element.getAttribute( 'min' ) ) ? this.element.getAttribute( 'min' ) : 0;
+        const max = isNumber( this.element.getAttribute( 'max' ) ) ? this.element.getAttribute( 'max' ) : 100;
+        const step = isNumber( this.element.getAttribute( 'step' ) ) ? this.element.getAttribute( 'step' ) : ( props.showScale ? 10 : 1 ); //( props.type === 'decimal' ? 0.1 : 1 );
+        props.min = Number( min );
+        props.max = Number( max );
+        props.step = Number( step );
         props.vertical = !props.appearances.includes( 'horizontal' );
         props.ticks = !props.appearances.includes( 'no-ticks' );
         props.maxTicks = 10;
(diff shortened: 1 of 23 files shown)


src/widget/analog-scale/analog-scalepicker.js
```diff
@@ -133,12 +133,14 @@ class AnalogScaleWidget extends RangeWidget {
     get props() {
         const props = this._props;
         props.touch = support.touch;
+        props.showScale = props.appearances.includes( 'show-scale' );
         props.min = isNumber( this.element.getAttribute( 'min' ) ) ? this.element.getAttribute( 'min' ) : 0;
         props.max = isNumber( this.element.getAttribute( 'max' ) ) ? this.element.getAttribute( 'max' ) : 100;
-        props.step = isNumber( this.element.getAttribute( 'step' ) ) ? this.element.getAttribute( 'step' ) : 1; //( props.type === 'decimal' ? 0.1 : 1 );
+        props.step = isNumber( this.element.getAttribute( 'step' ) ) ? this.element.getAttribute( 'step' ) : ( props.showScale ? 10 : 1 ); //( props.type === 'decimal' ? 0.1 : 1 );
```
MartijnR, line 139: This is your proxy "range" detection, right? Hmm, that seems quite bullet proof.

However, what if a user uses `range` without `step` (not mandatory) and with `show-scale`? (I haven't tested this)

src/widget/analog-scale/analog-scalepicker.js
```diff
@@ -133,12 +133,14 @@ class AnalogScaleWidget extends RangeWidget {
...
         props.vertical = !props.appearances.includes( 'horizontal' );
         props.ticks = !props.appearances.includes( 'no-ticks' );
         props.maxTicks = 10;
+        props.distress = props.showScale;
```
MartijnR, line 143: Ah I get it now. This functionality was more or less there already! :D Great you found that! 

I think we just have to change the name of this property (so also in the range widget) to make it more descriptive. Here is a test form that include the distress picker and the regular range widget to test for regressions: https://docs.google.com/spreadsheets/d/1KLQiQyQ5BlN_wd-83p8Eb6xIU-TZp-yvzgrm20HecRI/edit#gid=0. And here it is live: https://enke.to/widgets

How about `showScale` as you've already done?

gushil: Got it. Will do that.

src/widget/analog-scale/analog-scalepicker.js
gushil: Tested it and the widget will just show 'usual' range widget with no step-scale visible.

gushil:
Hi, @MartijnR I've already done some tests and I think this implementation is sufficient for the current requirements.
I plan to add a unit test for this issue. If you don't mind, maybe you can give me some hint about it. 

Thanks.

MartijnR:
Cool. I'll review after you added some tests.

Testing widgets is pretty cool now if I can say so myself. See https://github.com/enketo/enketo-core/blob/master/test/spec/widget.analog-scale.spec.js

And if you want to test things that are not part of helpers/test-widget, here is an example of how that could be done: https://github.com/enketo/enketo-core/blob/master/test/spec/widget.date-native.spec.js#L12-L21 (you'll find other examples in other widget.x.spec.js files as well.

gushil:
Hi @MartijnR, I've added some tests.

Thanks.

test/spec/widget.analog-scale.spec.js
```diff
@@ -10,3 +10,68 @@ const FORM2 = FORM1.replace( 'step="1"', 'step="0.1"' );
...
 runAllCommonWidgetTests( AnalogScaleWidget, FORM1, '2' );
 runAllCommonWidgetTests( AnalogScaleWidget, FORM2, '2.1' );
+
+const FORM_SHOW_SCALE =
+    `<label class="question non-select or-appearance-analog-scale or-appearance-vertical or-appearance-show-scale">
+        <span lang="" class="question-label active">Analog Scale</span>
+        <input type="number" name="/widgets/analog_scale/widget1" data-type-xml="int">
+    </label>`;
+
+testStaticProperties( AnalogScaleWidget );
```
MartijnR, line 20: `testStaticProperties` is already called (twice) with runAllCommonWidgetTests.

test/spec/widget.analog-scale.spec.js
```diff
@@ -10,3 +10,68 @@ const FORM2 = FORM1.replace( 'step="1"', 'step="0.1"' );
...
+    it( 'adds widget that contain 9 visible scales that not max and min', () => {
+        expect( widgetScalesContainer.querySelectorAll( '.range-widget__scale__between' ).length ).toEqual( 9 );
+    } );
+
+    it( 'adds widget that contain 9 visible scales that not max and min with the right values', () => {
+        let scalesNotMinMax = widgetScalesContainer.querySelectorAll( '.range-widget__scale__between' );
+        let scalesNotMinMaxValues = [];
+
+        Array.from( scalesNotMinMax ).forEach( ( scale ) => scalesNotMinMaxValues.push( parseInt( scale.textContent, 10 ) ) );
+        scalesNotMinMaxValues.sort();
```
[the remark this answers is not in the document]
gushil: To make sure that the values we got from all `'.range-widget__scale__between'` element are sorted.
Will try to test again with `sort` removed.

test/spec/widget.analog-scale.spec.js
gushil: Got it, will remove that.

test/spec/widget.analog-scale.spec.js
```diff
@@ -10,3 +10,68 @@ const FORM2 = FORM1.replace( 'step="1"', 'step="0.1"' );
...
+
+    it( 'adds widget that contain 9 visible scales that not max and min', () => {
+        expect( widgetScalesContainer.querySelectorAll( '.range-widget__scale__between' ).length ).toEqual( 9 );
+    } );
+
+    it( 'adds widget that contain 9 visible scales that not max and min with the right values', () => {
+        let scalesNotMinMax = widgetScalesContainer.querySelectorAll( '.range-widget__scale__between' );
+        let scalesNotMinMaxValues = [];
+
+        Array.from( scalesNotMinMax ).forEach( ( scale ) => scalesNotMinMaxValues.push( parseInt( scale.textContent, 10 ) ) );
```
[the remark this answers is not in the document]
gushil: Got it, will change to `.map`

gushil:
> There is an exception occurring when loading [this form](https://docs.google.com/spreadsheets/d/1C4cIy-PrhYTCFFawivlUvoQWF7ehdo1SIWbSLikdfRk/edit#gid=0)

The exception is now gone, but I have a question. In [the form](https://docs.google.com/spreadsheets/d/1C4cIy-PrhYTCFFawivlUvoQWF7ehdo1SIWbSLikdfRk/edit#gid=0) you mentioned above, I saw that when range input type, the comment said that **should ignore show-scale** with this appearances:
- analog-scale vertical show-scale
- analog-scale horizontal no-ticks show-scale
- analog-scale show-scale

What does it means? Should we not show labels?

Thanks

MartijnR:
> The exception is now gone, but I have a question. In the form you mentioned above, I saw that when range input type, the comment said that should ignore show-scale with this appearances:

Yes, I think so assuming that it won't work out of the box with custom min, max and step value.

MartijnR:
continued here: https://github.com/enketo/enketo-core/pull/739

PR #739, opened 2020-08-26
Feature/enketo express oc 359

MartijnR:
Hi Agus,

I thought it would be fun to switch roles. So please review the 3 commits I added to your branch, and feel free to continue working in this branch instead of your fork (and in the future also fine to work in github.com/enketo).

I did some minor refactoring of the tests and added the logic that completely ignores `show-scale` if used in conjunction with `horizontal` and/or `no-ticks`. I hope you find some errors this addition.

What is remaining (for you) is:

* [x] some css issue with a horizontal widget min/max values 
Screen Shot 2020-08-26 at 5 18 59 PM

* [x] some issue with location of reset button in horizontal widget, see above screenshot for that too

Master is like this:

Screen Shot 2020-08-26 at 5 15 21 PM

* [ ] ignoring show-scale for type=range, if feasible. I'm not actually sure if this is detectable. If not, I think we could just leave it.

### DIFF
--- a/src/widget/analog-scale/analog-scalepicker.js
+++ b/src/widget/analog-scale/analog-scalepicker.js
@@ -133,12 +133,14 @@ class AnalogScaleWidget extends RangeWidget {
     get props() {
         const props = this._props;
         props.touch = support.touch;
+        props.showScale = props.appearances.includes( 'show-scale' );
         props.min = isNumber( this.element.getAttribute( 'min' ) ) ? this.element.getAttribute( 'min' ) : 0;
         props.max = isNumber( this.element.getAttribute( 'max' ) ) ? this.element.getAttribute( 'max' ) : 100;
-        props.step = isNumber( this.element.getAttribute( 'step' ) ) ? this.element.getAttribute( 'step' ) : 1; //( props.type === 'decimal' ? 0.1 : 1 );
+        props.step = isNumber( this.element.getAttribute( 'step' ) ) ? this.element.getAttribute( 'step' ) : ( props.showScale ? 10 : 1 ); //( props.type === 'decimal' ? 0.1 : 1 );
         props.vertical = !props.appearances.includes( 'horizontal' );
         props.ticks = !props.appearances.includes( 'no-ticks' );
         props.maxTicks = 10;
+        props.distress = props.showScale;
 
         return props;
     }
--- a/src/widget/analog-scale/analog-scalepicker.js
+++ b/src/widget/analog-scale/analog-scalepicker.js
@@ -133,15 +133,15 @@ class AnalogScaleWidget extends RangeWidget {
     get props() {
         const props = this._props;
         props.touch = support.touch;
-        props.showScale = props.appearances.includes( 'show-scale' );
+        props.vertical = !props.appearances.includes( 'horizontal' );
+        props.ticks = !props.appearances.includes( 'no-ticks' );
+        props.showScale = props.appearances.includes( 'show-scale' ) && props.vertical && props.ticks && !this.element.type === 'range';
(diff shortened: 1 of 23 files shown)


MartijnR:
P.S. A nice demonstration that the tests you added are useful (see failure on one of my commits): https://travis-ci.org/github/enketo/enketo-core/builds

src/widget/analog-scale/analog-scalepicker.js
```diff
@@ -133,11 +133,15 @@ class AnalogScaleWidget extends RangeWidget {
     get props() {
         const props = this._props;
         props.touch = support.touch;
-        props.min = isNumber( this.element.getAttribute( 'min' ) ) ? this.element.getAttribute( 'min' ) : 0;
-        props.max = isNumber( this.element.getAttribute( 'max' ) ) ? this.element.getAttribute( 'max' ) : 100;
-        props.step = isNumber( this.element.getAttribute( 'step' ) ) ? this.element.getAttribute( 'step' ) : 1; //( props.type === 'decimal' ? 0.1 : 1 );
         props.vertical = !props.appearances.includes( 'horizontal' );
         props.ticks = !props.appearances.includes( 'no-ticks' );
+        props.showScale = props.appearances.includes( 'show-scale' ) && props.vertical && props.ticks;
```
gushil, line 138: Got it. So this is how we restrict this feature to vertical style, with show-scale, and with no-ticks.

test/spec/widget.analog-scale.spec.js
```diff
@@ -10,3 +10,56 @@ const FORM2 = FORM1.replace( 'step="1"', 'step="0.1"' );
...
+
+testBasicInstantiation( AnalogScaleWidget, FORM_SHOW_SCALE );
+
+describe( 'Analog-scale widget with show scale', () => {
+
+    const fragment = document.createRange().createContextualFragment( FORM_SHOW_SCALE );
+    new AnalogScaleWidget( fragment.querySelector( 'input' ) );
+    const widget = fragment.querySelector( '.range-widget' );
+    const widgetInput = widget.querySelector( 'input' );
+    const widgetScalesContainer = widget.querySelector( '.range-widget__scale' );
```
gushil, line 28: Interesting. So we don't need `beforeEach` hook because there is no change in the widget we are testing. Additionally, it made the test run longer because we have to set up these variables every time the test (`it`) runs.

gushil: Got it. We use `const` because the variables will not be changed in these tests.

gushil (commented):
Thanks @MartijnR 
Learn a lot from your changes!

gushil:
@MartijnR 
I've looked into the HTML produced by enketo-transformer, it looks like that the `input` produced for question type `range` and `integer` is similar, we have nothing to differentiate both questions.

Should we look (and probably make some changes) into enketo-transformer further?

src/widget/analog-scale/analog-scalepicker.js
gushil: Hmm, I'm not sure what error should I look into with this line. It looks like that all requirement is fulfilled.
Actually, I become more understanding of what the ticket says by looking into this line.

MartijnR: Oh sorry, I didn't leave an error deliberately! :) I was just encouraging a thorough look!

MartijnR:
> Should we look (and probably make some changes) into enketo-transformer further?

Yes, that would indeed be the approach to use. However, let's just leave it for now since OC hasn't published the range type yet.

MartijnR:
Great!

PR #740, opened 2020-08-28
Feature/enketo express oc 370

gushil:
close https://github.com/OpenClinica/enketo-express-oc/issues/370

### DIFF
--- a/src/widget/analog-scale/analog-scalepicker.scss
+++ b/src/widget/analog-scale/analog-scalepicker.scss
@@ -210,3 +210,7 @@ $q-pad-right: 0px !default;
         @include display-flex;
     }
 }
+
+.pages.or [role="page"]:not(.current) {
+    display: none;
+}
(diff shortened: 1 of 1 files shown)


src/widget/analog-scale/analog-scalepicker.scss
```diff
@@ -210,3 +210,7 @@ $q-pad-right: 0px !default;
         @include display-flex;
     }
 }
+
+.pages.or [role="page"]:not(.current) {
```
MartijnR, line 214: This rule is not specific to analog-scale widgets. I think the fix is good, but it should be made around here I think: https://github.com/enketo/enketo-core/blob/master/src/sass/core/_pages.scss#L17

Alternatively, perhaps instead of this change, perhaps a different kind of change in analog-scalepicker.scss could be made to make sure the current pages.scss rule 'wins' (I'm not sure how).

gushil: Got it. I think I know where the specific changes should be made.

Issue #363, opened 2016-02-24
Travis PhantomJS error


MartijnR:
Cannot reproduce locally.

Maybe relevant: https://github.com/travis-ci/travis-ci/issues/3225

MartijnR:
The issue is with MergeXML it seems (line 500).

PR #744, opened 2020-10-01
fixed: PDF endpoint creates image-map that overlaps in cell below with Grid theme

theywa:
[#363](https://github.com/OpenClinica/enketo-express-oc/issues/363)

### DIFF
--- a/src/js/print.js
+++ b/src/js/print.js
@@ -204,8 +204,12 @@ function _resizeRowElements( row, maxWidth ) {
     } );
 
     row.forEach( el => {
-        el.classList.add( 'print-height-adjusted' );
-        el.style.height = `${maxHeight}px`;
+        // unset max height for image-map widget 
+        // (https://github.com/OpenClinica/enketo-express-oc/issues/363)
+        if ( !el.classList.contains( 'or-appearance-image-map' ) ) {
+            el.classList.add( 'print-height-adjusted' );
+            el.style.height = `${maxHeight}px`;
+        }
     } );
 }
 


src/widget/image-map/image-map.js
```diff
@@ -254,6 +257,19 @@ class ImageMap extends Widget {
...
+    /**
+     * If somehow the image not fully loaded before applying print style
+     * Then we need to recalculate the height
+     */
+    _recalculateHeight() {
+        const questionHeight = this.question.offsetHeight;
+        const fieldsetHeight = this.question.querySelector( 'fieldset' ).offsetHeight;
+        if ( this.question.classList.contains( 'print-height-adjusted' ) && ( fieldsetHeight > questionHeight ) ) {
+            // 19 px got from expected height - currentOffsetHeight
+            this.question.style.height = `${fieldsetHeight + 19}px`;
```
MartijnR, line 269: I don't quite understand the 19px (after reading comment). Can you get this value programmatically?

MartijnR (commented):
Hi @theywa,

Would handling the printify event be a more targeted solution? https://github.com/enketo/enketo-core/blob/614bb46c7878c5a3b96c7ec5eeed641d12b33504/src/widget/text-print/text-print.js#L20

theywa:
Hi @MartijnR 
In process of handling it from text-print, I found that unset the max height for image-map widget solved the height issue so the code is updated. Do you think is safe to unset the max height cell for image-map?

MartijnR:
Quite safe probably but not sure. You may want to try a form with a (w4) row that has an image map widget (w3) as well as a regular text input or something (w1), and see if both cells are stretched to full row height. If that works, that fix sounds great with me (until somebody reports an issue).

theywa:
@MartijnR ~ sorry for the late response, here is the screenshot from image map widget(w3) and text input(w1)
Screenshot from 2020-10-08 19-43-42
it looks like the input(w1) have a height problem there

Issue #745, opened 2020-10-10
Safari: Date-time calculations appear to include UTC offsets


nanzhang01:
**Describe the bug**
When a datetime question has a calculation, the result appears to be adjusted by the UTC offset.

**Sample xls form**
[date-time.calc.has.utc.offset.xlsx](https://github.com/enketo/enketo-core/files/5358618/date-time.calc.has.utc.offset.xlsx)
 
**Screenshots**
Calculation `now() + 0` produces a result +8 hours in China Standard Time.
Chrome:
![Screen Shot 2020-10-10 at 3 56 50 PM](https://user-images.githubusercontent.com/29697780/95649766-c5741200-0b11-11eb-9af7-21e77c1aada9.png)

Safari on MacOS:
<img width="608" alt="Screen Shot 2020-10-10 at 3 55 58 PM" src="https://user-images.githubusercontent.com/29697780/95649768-cc028980-0b11-11eb-883e-19f8cb89b86c.png">

 
**Browser and OS (please complete the following information):**
 - Only Safari on macOS can reproduce this issue and other browsers work as expected.

nanzhang01:
@MartijnR, The returned calculation result is correct from xpathJS, the key is: the native Date constructor returns different values between Safari and other browsers. I would like to create a PR to fix this.

MartijnR:
Thanks for reporting, and yes please create a PR 👍

PR #747, opened 2020-10-22
Fix: let the Date constructor return the correct time zone on safari.

nanzhang01:
The fix for #745.


src/js/sniffer.js
```diff
@@ -26,6 +26,58 @@ const os = {
      **/
     get android() {
         return /android/i.test( ua );
+    },
+
```
MartijnR, line 30: Maybe a bit nitpicky, but I'd like a separate browser object. We're trying to avoid sniffing and remove any sniffing code when we don't use it.

```
const browser = { get safari(){ return ua.includes('Safari')}
```

src/js/format.js
```diff
@@ -75,6 +75,29 @@ const time = {
      */
     hasMeridian( time ) {
         return HAS_MERIDIAN.test( _cleanSpecialChars( time ) );
+    },
+    /**
+     * getTimezoneOffsetAsTime
+     */
+    getTimezoneOffsetAsTime() {
```
MartijnR, line 82: This function has already been added to the Date prototype (not great practice but it's there): https://github.com/enketo/enketo-xpathjs/blob/master/src/date-extensions.js#L23, (so new Date().getTimezoneOffsetAsTime() I think) or this there a subtle difference?

MartijnR (changes requested):
Thank you!

src/js/sniffer.js
```diff
@@ -26,6 +26,13 @@ const os = {
...
+    /**ua
+     * @type {string}
+     **/
+    get safari() {
+        return /^((?!chrome|android).)*safari/i.test( ua );
```
MartijnR, line 35: I don't understand the negative chrome|android lookahead. After reading the issue again, I realize you probably need to add an "os.macos" and a "browser.safari" sniffer right?

MartijnR (changes requested):

src/widget/datetime/datetimepicker-extended.js
```diff
@@ -121,8 +122,17 @@ class DatetimepickerExtended extends Widget {
...
+        let val = ( $dateTimeI.val().length > 0 ) ? new Date( $dateTimeI.val() ).toISOLocalString() : '';
+        /**
+         * fix a bug which is only on safari (#745)
+         * If the local timezone is +08:00, for a date value of new Date('2020-10-10T13:10:10') will be:
+         * *** in chrome, the value is: Sat Oct 10 2020 13:10:10 GMT+0800 (China Standard Time)
+         * *** but in safari, the value is: Sat Oct 10 2020 21:10:10 GMT+0800 (CST)
+         * so we have to append the timezone here
+         */
+        if ( os.safari ) {
+            val = ( $dateTimeI.val().length > 0 ) ? ( new Date( $dateTimeI.val() + new Date( dt ).timezoneOffsetAsTime ) ).toISOLocalString() : '';
```
MartijnR, line 134: `dt` is not defined (see also Travis error in PR)

src/js/sniffer.js
nanzhang01: We can't detect the safari browser by only check the string: `safari`  from `navigator.userAgent` .  because  `safari` exists in the userAgent for Chrome/Android, too. That's why I have to exceeded `chrome` and `android` here.
image

src/widget/datetime/datetimepicker-extended.js
nanzhang01: Thanks, fixed now.

src/js/sniffer.js
MartijnR: Oh, right. Thanks

Issue #748, opened 2020-11-16
Label not being displayed for calculate field using jr:choice-name in repeat-group


FrankApiyo:
**Describe the bug**
calculate field label that is calculated using `jr:choice-name` is not displayed as expected for the following form
[jr_choice_name_repeats.xlsx](https://github.com/enketo/enketo-core/files/5546093/jr_choice_name_repeats.xlsx)

**To Reproduce**
- open the above form in enketo
- Click on any option in the `province` question
- note that the label `counties in` is missing the label that is supposed to postfix it

**Screenshots**
- current behaviour
![jr_choice_name-in-repeats-groups](https://user-images.githubusercontent.com/25260439/99251428-62ffe700-281e-11eb-8292-35ff8f0dd2cf.png)
- expected behaviour
![Enketo-Core-Developer-template](https://user-images.githubusercontent.com/25260439/99251480-77dc7a80-281e-11eb-9d68-2ba85b09b1e3.png)

**Browser and OS (please complete the following information):**
 - OS: Ubuntu 18
 - Browser: Chrome

**Additional context**
The `/data/r1/province_name` seems to be missing a `name` attribute
![Screenshot from 2020-11-16 15-16-10](https://user-images.githubusercontent.com/25260439/99251940-38625e00-281f-11eb-9efa-b63ea2ec8dd0.png)

PR #749, opened 2020-11-16
Fix/issue 748

FrankApiyo:
Fixes: #748

### DIFF
--- a/src/js/form.js
+++ b/src/js/form.js
@@ -413,7 +413,10 @@ Form.prototype.replaceChoiceNameFn = function( expr, resTypeStr, context, index,
             let label = '';
             const value = this.model.evaluate( params[ 0 ], resTypeStr, context, index, tryNative );
             const name = stripQuotes( params[ 1 ] ).trim();
-            const inputs = [ ...this.view.html.querySelectorAll( `[name="${name.startsWith( '/' ) ? name : joinPath( context, name )}"]` ) ];
+            var inputs = [ ...this.view.html.querySelectorAll( `[name="${name.startsWith( '/' ) ? name : joinPath( context, name )}"]` ) ];
+            inputs = inputs.length ?
+                    inputs :
+                    [ ...this.view.html.querySelectorAll( `[data-name="${name.startsWith( '/' ) ? name : joinPath( context, name )}"]` ) ];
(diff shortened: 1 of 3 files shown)


src/js/form.js
```diff
@@ -413,7 +413,10 @@ Form.prototype.replaceChoiceNameFn = function( expr, resTypeStr, context, index,
             let label = '';
             const value = this.model.evaluate( params[ 0 ], resTypeStr, context, index, tryNative );
             const name = stripQuotes( params[ 1 ] ).trim();
-            const inputs = [ ...this.view.html.querySelectorAll( `[name="${name.startsWith( '/' ) ? name : joinPath( context, name )}"]` ) ];
+            var inputs = [ ...this.view.html.querySelectorAll( `[name="${name.startsWith( '/' ) ? name : joinPath( context, name )}"]` ) ];
+            inputs = inputs.length ?
+                    inputs :
+                    [ ...this.view.html.querySelectorAll( `[data-name="${name.startsWith( '/' ) ? name : joinPath( context, name )}"]` ) ];
```
MartijnR, line 419: Nice work!

I'd slightly prefer searching directly for `[name="${name}"],[data-name="${name}"]` in one go (and update the name variable beforehand).

Some background:

Only radio buttons don't use the path in the `name` attribute. The reason is that this would make all repeated questions with radiobuttons behave as one. Hence, we use the unfortunate `data-name` attribute for radiobuttons. In hindsight, it would have been better to not use name at all but use data-name (or better: data-path) for everything.

MartijnR (changes requested):
As you probably feared, could you also please add a test? 😃 

Thankfully, there are some examples already: https://github.com/enketo/enketo-core/blob/master/test/spec/form.spec.js#L1133

Please simplify the XForm you added as much as possible (2 options?) to use for your test.

Note, we're slowly removing jQuery and form.view.$, so use form.view.html instead please.

FrankApiyo:
Please let me know if I should squash this into a single commit, this is ready for another review. Very excited to have a PR merged here :smile:

MartijnR:
Will check it out! It always fine to have a dirty commit history in a PR. It will always be squashed when merging.

MartijnR (approved):
Excellent!

PR #754, opened 2020-12-01
fixed: primary instance node names starting with underscore followed …

MartijnR:
…by number, break autocomplete widget, closes 752,

thanks @nanzhang01!


MartijnR:
@nanzhang01, would you mind reviewing this PR please? Thanks for the solution! Maybe there are places you found, that I missed.

(P.S. I added you to the contributors team, so once you accept, I can formally request a review from you via the interface).

nanzhang01 (approved):

Issue #378, opened 2016-07-26
Improve performance of pull-data()


MartijnR:
Confirmed: first item in list takes just as long as last item in list. So there is scope to improve with factor 2 on average.

MartijnR:
see https://github.com/kobotoolbox/enketo-express/issues/534

MartijnR:
* [ ] get rid of tryNative completely?
* [ ] look at Alex' XPath evaluator. Possible to get all XPathJS tests to pass?

PR #760, opened 2021-02-05
update: query selector to target element with '.option-wrapper' only

theywa:
[#378](https://github.com/OpenClinica/enketo-express-oc/issues/378)

### DIFF
--- a/src/widget/image-map/image-map.js
+++ b/src/widget/image-map/image-map.js
@@ -95,7 +95,7 @@ class ImageMap extends Widget {
 
                     // remove images in all languages
                     that.question.querySelectorAll( 'img' ).forEach( el => el.remove() );
-                    that.question.querySelector( 'div[class=option-wrapper]' ).before( fragment );
+                    that.question.querySelector( 'fieldset > .option-wrapper' ).before( fragment );
                     const widget = that.question.querySelector( '.image-map' );
                     const svg = widget.querySelector( 'svg' );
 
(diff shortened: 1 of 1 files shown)


src/widget/image-map/image-map.js
```diff
@@ -95,7 +95,7 @@ class ImageMap extends Widget {
 
                     // remove images in all languages
                     that.question.querySelectorAll( 'img' ).forEach( el => el.remove() );
-                    that.question.querySelector( '.option-wrapper' ).before( fragment );
+                    that.question.querySelector( 'div[class=option-wrapper]' ).before( fragment );
```
MartijnR, line 98: Great find! 

This solution would be fragile, as it would break if we ever add another class to form option-wrapper. I'd prefer something like `('*:not(.widget) .option-wrapper')`.

MartijnR (changes requested):

src/widget/image-map/image-map.js
theywa: test the selector above but the result still the same
new selector 1

since the image will be attached on the fieldset node, is it okay to use something like `('fieldset:not(.widget) > .option-wrapper')` or `('fieldset > .option-wrapper')` ?
new selector 2

MartijnR: Thanks for checking! I think I understand why my suggestion doesn't work. It would be great to use the `widget` class to exclude it though...

Hmm... `[...document.querySelectorAll('.option-wrapper')].filter(el => !el.closest('.widget'))[0]` is maybe not very elegant.

Ok, let's just do `('fieldset > .option-wrapper')` then.

theywa: no problem and code updated.

MartijnR (approved):

Issue #456, opened 2017-09-11
Missing tag for 4.38.2


alxndrsn:
Possibly 75fb69caf9024f07be789d8d7874dfa82aca8c22

MartijnR:
thanks! it's added now.

PR #763, opened 2021-02-26
fixed: Issue With Printing Single Select Items with Minimal Display on Chrome and Edge

theywa:
[#456](https://github.com/OpenClinica/enketo-express-oc/issues/456)


MartijnR (changes requested):
will break cell height stretch feature with questions next to minimal question (on same row)

MartijnR:
will break cell height stretch feature with questions next to minimal question (on same row)

PR #768, opened 2021-03-11
Improve accessibility

### LINKED ISSUE #137: screen reader support
from https://community.kobotoolbox.org/t/accessibility-standards-for-visually-disabilities/10885

When filling out the form using a screen reader programme:

* The voice-over reads content that is not actually visible on the page. I assume the programme reads something from the code of the page, including content from other pages. The voice-over should read only the content that is visible on the page.
* When moving to the next page, screen reader by default goes to the bottom of a page. It is very inconvenient when a user with visual impairment must manually search and find the beginning of a page. When clicking “next”, it should take you to the next question by default.

### LINKED ISSUE #735: screen reader support
from https://community.kobotoolbox.org/t/accessibility-standards-for-visually-disabilities/10885

When filling out the form using a screen reader programme:

* The voice-over reads content that is not actually visible on the page. I assume the programme reads something from the code of the page, including content from other pages. The voice-over should read only the content that is visible on the page.
* When moving to the next page, screen reader by default goes to the bottom of a page. It is very inconvenient when a user with visual impairment must manually search and find the beginning of a page. When clicking “next”, it should take you to the next question by default.

FrankApiyo:
closes enketo/enketo#137

**NB:** These changes were tested manually on chrome using the *[chromevox](https://chrome.google.com/webstore/detail/screen-reader/kgejglhpjiefppelpmljglcjbhoiplfn?hl=en)* screen reader/extension

### DIFF
--- a/src/js/form.js
+++ b/src/js/form.js
@@ -790,6 +790,7 @@ Form.prototype.setEventHandlers = function() {
         // update the form progress status
         this.progress.update( event.target );
     } );
+
     this.view.html.addEventListener( events.FakeFocus().type, event => {
         // update the form progress status
         this.progress.update( event.target );
--- a/src/js/page.js
+++ b/src/js/page.js
@@ -52,6 +52,7 @@ export default {
                 this.$btnFirst = this.$formFooter.find( '.first-page' );
                 this.$btnPrev = this.$formFooter.find( '.previous-page' );
                 this.$btnNext = this.$formFooter.find( '.next-page' );
+                this.$btnNext.attr( 'tabindex', 2 );
                 this.$btnLast = this.$formFooter.find( '.last-page' );
                 this.$toc = $( formWrapper.querySelector( '.pages-toc__list' ) );
                 this._updateAllActive( allPages );
@@ -367,6 +368,7 @@ export default {
             this._focusOnFirstQuestion( pageEl );
             this._toggleButtons( newIndex );
             pageEl.dispatchEvent( events.PageFlip() );
+            pageEl.setAttribute( 'tabindex', 1 );
         }
     },
     /**
@@ -398,6 +400,9 @@ export default {
             .find( 'input, select, textarea' )
             .eq( 0 )
             .trigger( 'fakefocus' );
+
+        // focus on element
+        pageEl.focus();
 
         pageEl.scrollIntoView();
     },
--- a/src/js/readonly.js
+++ b/src/js/readonly.js
@@ -20,6 +20,12 @@ export default {
             const empty = !node.value && !node.dataset.calculate && !setValue && !node.classList.contains( 'readonly-forced' );
 
             node.classList.toggle( 'empty', empty );
+
+            if( empty ){
+                node.setAttribute( 'aria-hidden', 'true' );
+            }else{
+                node.removeAttribute( 'aria-hidden' );
+            }
         } );
     }
 };


MartijnR (changes requested):
Thanks @FrankApiyo!

src/js/page.js
```diff
@@ -367,7 +367,17 @@ export default {
...
+            pageEl.focus();
         }
+
+        document.querySelectorAll( '[tabindex="0"]' ).forEach( el => el.setAttribute( 'tabIndex', -1 ) );
+        const nextPageButton = document.querySelector( '.next-page' );
+        if ( nextPageButton ) {
+            nextPageButton.setAttribute( 'tabIndex', 2 );
+        }
+        document.querySelectorAll( 'input[readonly]' ).forEach( el => el.setAttribute( 'aria-hidden', 'true' ) );
+        document.querySelectorAll( '[style*="display: none"]' ).forEach( el => el.setAttribute( 'aria-hidden', 'true' ) );
```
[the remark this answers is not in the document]
MartijnR: Actually, should readonly questions be hidden at all? E.g. a readonly question with a default value or calculated value?

Perhaps you'd like to only hide so-called empty readonly questions without calculations or dynamic defaults? Then definitely look at readonly.js to add this there.

src/js/page.js
```diff
@@ -367,7 +367,17 @@ export default {
             this._focusOnFirstQuestion( pageEl );
             this._toggleButtons( newIndex );
             pageEl.dispatchEvent( events.PageFlip() );
+            pageEl.setAttribute( 'tabIndex', 1 );
+            pageEl.focus();
```
MartijnR, line 371: Did you look into why `_focusOnFirstQuestion` is not working?

MartijnR (changes requested):

src/js/page.js
```diff
@@ -367,7 +367,17 @@ export default {
             this._focusOnFirstQuestion( pageEl );
             this._toggleButtons( newIndex );
             pageEl.dispatchEvent( events.PageFlip() );
+            pageEl.setAttribute( 'tabIndex', 1 );
```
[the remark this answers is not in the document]
FrankApiyo: oversight! fixing this

src/js/page.js
FrankApiyo: It seems that `_focusOnFirstQuestion` calls `trigger( 'fakefocus' );` .  After looking around for a while, it seems `fakefocus` is not handled anywhere.

src/js/page.js
```diff
@@ -367,7 +367,17 @@ export default {
...
+            pageEl.focus();
         }
+
+        document.querySelectorAll( '[tabindex="0"]' ).forEach( el => el.setAttribute( 'tabIndex', -1 ) );
+        const nextPageButton = document.querySelector( '.next-page' );
+        if ( nextPageButton ) {
+            nextPageButton.setAttribute( 'tabIndex', 2 );
+        }
```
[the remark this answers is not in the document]
FrankApiyo: going to move it to `init()`

src/js/readonly.js
```diff
@@ -12,6 +12,7 @@ export default {
         const nodes = this.form.getRelatedNodes( 'readonly', '', updated ).get();
         nodes.forEach( node =>  {
             node.closest( '.question' ).classList.add( 'readonly' );
+            node.setAttribute( 'aria-hidden', 'true' );
```
MartijnR, line 15: Just repeating my earlier question. Should all readonly inputs really be aria-hidden? It's very common to show a calculated value in a readonly input field. I just want to make sure.

src/js/page.js
```diff
@@ -29,6 +29,7 @@ export default {
      * @type {Function}
      */
     init() {
+        document.querySelectorAll( '[tabindex="0"]' ).forEach( el => el.setAttribute( 'tabIndex', -1 ) );
```
MartijnR, line 32: I don't feed good about this kind of DOM correction. Makes more sense to fix at the source, or am I overlooking something?

src/js/page.js
```diff
@@ -65,6 +66,11 @@ export default {
                 this._setLangChangeHandlers();
                 this.active = true;
                 this._flipToFirst();
+
+                const nextPageButton = document.querySelector( '.next-page' );
```
MartijnR, line 70: see $btnNext a few lines above

src/js/page.js
MartijnR: Well, it seems to me that focusOnFirstQuestion's purpose should be to focus on the first question... :)  So it makes sense to me to fix that instead. 

page.js should actually be de-jquerified (eventually). Listening should be done with `events.FakeFocus().type` and triggering with `events.FakeFocus()`. It looks like it's used to update the progress bar only.

Surprisingly this function only appears to be scrolling to the first question and not focusing at all. I don't remember why, but let's add the focus there! 

Note that a full-page form also doesn't focus on the first question. Is that also a problem with screen readers?

MartijnR (changes requested):

lognaturel:
> Note that a full-page form also doesn't focus on the first question. Is that also a problem with screen readers?

Forgive me, I'm jumping in here without a lot of context but this happened to catch my eye. I've found this slightly annoying myself. On a computer pressing tab once gets the expected focus but that's not very discoverable. On mobile it would be very nice to have the keyboard out and ready to go when starting a new record.

A related issue that I find even more problematic is what happens after submission in a non-single context. We're left at the bottom of the screen and have to manually scroll up and put focus on the first question to start a new record. Forms can have hundreds or thousands of questions so this can add a lot of friction. Even knowing this behavior I get confused so I expect it's an accessibility issue. I would expect that after a successful submission, the page scrolls to the top and the first question has focus. If this is worth following up on but doesn't feel related to this work, I'm happy to file a separate issue.

MartijnR:
thanks @lognaturel!

> e left at the bottom of the screen and have to manually scroll up and put focus on the first question 

Interestingly, I never realized that and am not sure if this ever worked! I've added an issue here just in case, for scrolling: https://github.com/enketo/enketo-express/issues/248 (the scrolling likely needs a code change in enketo-express).

Wrt to focusing on the first question, I hadn't thought of the keyboard popping up on mobile browsers. I would have thought that undesirable (and that's maybe why we don't have this). So good to know that seems like a helpful thing to you, and I'd have no problem introducing this. We can always easily remove it again if it causes issues.

So if you agree @FrankApiyo, and if it is within your scope, please focus a one-page form on the first question as well.

src/js/readonly.js
FrankApiyo: I'm not certain if all read-only fields are hidden, but it seems those with *class* `empty` are hidden for sure https://github.com/enketo/enketo-core/blob/78a971943c2e1b4d7c85c618a6522c8be538941e/src/sass/core/_main.scss#L297

going to change this to check if *class* `empty` is present in the node

src/js/page.js
FrankApiyo: it seems this line isn't needed after all; after testing these changes again with a screen reader, it seems focus stays were it's expected to without this line

src/js/page.js
FrankApiyo: `$btnNext` seemed to be missing the `setAttribute` method

FrankApiyo: going to use `attr` method

src/js/page.js
FrankApiyo: moving the call to `.focus()` to the *'fakefocus'* event handler makes sense. however, I'm not sure why the event handler is not called at all in spite of the even being fired. this is happening for me locally.

MartijnR (changes requested):

src/js/readonly.js
```diff
@@ -20,6 +20,10 @@ export default {
             const empty = !node.value && !node.dataset.calculate && !setValue && !node.classList.contains( 'readonly-forced' );
 
             node.classList.toggle( 'empty', empty );
+
+            if( node.classList.contains( 'empty' ) ){
+                node.setAttribute( 'aria-hidden', 'true' );
+            }
```
MartijnR, line 26: Doesn't this mean the attribute will remain there forever (even if it becomes non-empty)?

Also `if(empty)` would probably be faster.

src/js/form.js
```diff
@@ -790,7 +790,11 @@ Form.prototype.setEventHandlers = function() {
         // update the form progress status
         this.progress.update( event.target );
     } );
+
     this.view.html.addEventListener( events.FakeFocus().type, event => {
+        // focus on element
+        event.target.focus();
+
```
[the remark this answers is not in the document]
MartijnR: Can this be removed now, that you've moved it somewhere else?

MartijnR (changes requested):

src/js/readonly.js
FrankApiyo: going to remove `aria-hidden`  attribute when not `empty` is false

MartijnR (approved):

MartijnR (approved):
Thanks!

PR #770, opened 2021-03-17
feature: use v2 enketo geocoder api

green3g:
https://github.com/enketo/enketo-express/issues/123

Utilize enketo-express geocoder api `/api/v2/geocoder`

To test this, I set up a simple proxy to enable communication between ee and core:

Proxy: localhost:3000
Enketo: localhost: 8006 (via config)
Core: localhost:8005

```js
const express = require( 'express' );
const { createProxyMiddleware } = require( 'http-proxy-middleware' );

const PORT = 3000;
const app = express();
app.use( '/api', createProxyMiddleware( { target: 'http://localhost:8006', changeOrigin: true } ) );
app.use( '/', createProxyMiddleware( { target: 'http://localhost:8005', changeOrigin: true } ) );
app.listen(  PORT );
console.log( `Proxy is listening on port http://localhost:${PORT}/` );
```


MartijnR:
Thanks! We should keep the google code for the map layers I think, and only change geocoding code.

roemhildtg:
Hi @MartijnR  - I apologize in the delay in getting these fixed - I have fallen behind due to some other priorities at the moment. I will be following up though in May. 

Just wanted to provide an update

Issue #463, opened 2017-10-12
time format incorrect


MartijnR:
Should be like `<start_time>09:15:00.000+03</start_time>` instead of  `<start_time>09:15</start_time>`.

See https://opendatakit.github.io/xforms-spec/#data-types

@alxndrsn, @spatialdude, @tedrick, this could be significant for you since you are not using ODK Collect (which is currently using the correct format). It will be corrected in Enketo Core shortly.

MartijnR:
* [ ] fix in XPath evaluator
* [ ] test time comparison in XPath evaluator
* [ ] change type in Enketo Core
* [ ] change widget in Enketo Core
* [ ] test loading old format from existing record

tedrick:
@zhif7782 @huoyixin

zhif7782:
@tedrick @MartijnR as far as you know, will there be any risk if we upgrade Enketo Core after the fix? Such as calculation will of time will be wrong or submitted results will be incorrect.

MartijnR:
@zhif7782 XPath expressions using time will be tested for issue with that will hopefully be avoided (at least time comparisons). I don't think we can do any calculations with time at the moment (and that won't change). In addition, the old time format will also still have to work when an old record is loaded into a form as loading data is a much-used Enketo feature.

The main thing is probably to just be able to handle the new format in the Survey123 backend or convert it to the old format before sending it through.

zhif7782:
Thanks @MartijnR for the explanation, will test accordingly after the issue has been fixed.

MartijnR:
Took me a while to figure out but it turns there has never been any support for time comparisons in Enketo (and ODK Collect) (as well as calculations), so there was no need to worry about this as that remains unchanged. All times are converted to number `NaN` when comparing (just like "a" < "b" is `false` and "b" < "a" is also `false`) in any compliant XPath evaluator.

PR #774, opened 2021-03-30
fixed: Form Printing adding page breaks in the middle of an item

theywa:
[#463](https://github.com/OpenClinica/enketo-express-oc/issues/463)

Using this XML [PostSurgery.xml.txt](https://github.com/enketo/enketo-core/files/6229185/PostSurgery.xml.txt) for testing also added `break-inside: avoid;` to make sure the header(title) move to the next pages too(when page breaks triggered).

### DIFF
--- a/src/sass/grid/_print.scss
+++ b/src/sass/grid/_print.scss
@@ -225,3 +225,15 @@ screen and (min-width: 700px) {
 }
 
 // End of adjustment to get page breaks to work in multi-line flexbox
+
+// Another workaround to get page breaks to works
+// https://github.com/OpenClinica/enketo-express-oc/issues/463
+.or {
+    &:not(.print-relevant-only) section.question.or-branch.disabled, 
+    &:not(.print-relevant-only) section.or-branch.disabled,
+    &:not(.print-relevant-only) section.question.or-branch:not(.disabled):not(.pre-init), 
+    &:not(.print-relevant-only) section.or-branch:not(.disabled):not(.pre-init) {
+        display: table;
+        break-inside: avoid;
+    }
+}


src/sass/grid/_print.scss
```diff
@@ -225,3 +225,13 @@ screen and (min-width: 700px) {
 }
 
 // End of adjustment to get page breaks to work in multi-line flexbox
+
+// Another workaround to get page breaks to works
+// https://github.com/OpenClinica/enketo-express-oc/issues/463
+.or {
+    &:not(.print-relevant-only) section.question.or-branch.disabled, 
+    &:not(.print-relevant-only) section.or-branch.disabled {
+        display: table;
```
MartijnR, line 234: Why `display: table` if we don't use that for any other questions, e.g. non-disabled questions. Why not use the same solution for everything? I cannot think of any reason why we would use a different solution for disabled vs. non-disabled.

MartijnR (changes requested):

src/sass/grid/_print.scss
MartijnR: We still have some custom solution for branches. It makes no sense to do this just for branches. I guarantee that either:

a. the bug still exists for other sections/question that are not a branch (have no relevant), or
b. that `display: table` is not necessary

But I'm getting tired of this now, and OC is waiting, so I'll just approve and merge this.

MartijnR (approved):

src/sass/grid/_print.scss
theywa: I'm sorry for the inconvenience

Issue #771, opened 2021-03-18
Range value doesn't display when draft (offline) survey is loaded


nmandrescu:
**Describe the bug**
Range value doesn't display when draft (offline) survey is loaded.

**To Reproduce**
[Range test.xlsx.zip](https://github.com/enketo/enketo-core/files/6164522/Range.test.xlsx.zip)
1. Create the form based on attached sample.
2. Open the survey in offline mode.
3. Configure some range value (e.g. 2).
4. Save as draft
5. Reopen the draft version
6. Notice the tick correctly shows the range value, but the value doesn't display.

**Expected behavior**
The range value must be displayed when form is loaded from draft.

**Screenshots**
Value displayed on first time selection (before saved as draft):
![selected](https://user-images.githubusercontent.com/17854915/111636422-23afca00-8801-11eb-8d9b-d1a7a67d0b14.png)
Value is gone if you reload the draft:
![reloaded-from-draft](https://user-images.githubusercontent.com/17854915/111636458-2dd1c880-8801-11eb-85f4-766bc079056e.png)

**Browser and OS (please complete the following information):**
 - OS: Linux
 - Browsers:
   - Chrome 89.0.4389.90 (Official Build) (64-bit)
   - Firefox 86.0.1 (64-bit)
   - Opera 74.0.3911.218

**Additional context**
1. In loaded draft version, clicking on the same value doesn't bring its display. You have to select another value (e.g. 3), and then select the initial one (e.g. 2) to see it displayed.
2. I have been using the online Kobo Toolbox that is based on official Enketo Express.
3. Here is a link for test https://ee.kobotoolbox.org/x/xCi9Grlc

MartijnR:
Thank you!

nmandrescu:
Please note the same issue occurs when opening the submission instance for edit.

PR #776, opened 2021-04-27
fixed: Range value doesn't display when draft (offline) survey is loaded

eyelidlessness:
Addresses #771.

I also updated the Node version in the development docs, if this is incorrect/unwanted I can revert. And I pinned the version for [Volta](https://volta.sh/) users in package.json, which again I can revert if unwanted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-![coverage-shield-badge-1](https://img.shields.io/badge/coverage-74.7%25-yellow.svg)
+![coverage-shield-badge-1](https://img.shields.io/badge/coverage-74.92%25-yellow.svg)
 [![npm version](https://badge.fury.io/js/enketo-core.svg)](http://badge.fury.io/js/enketo-core) [![Build Status](https://travis-ci.com/enketo/enketo-core.svg?branch=master)](https://travis-ci.org/enketo/enketo-core) [![Dependency Status](https://david-dm.org/enketo/enketo-core/status.svg)](https://david-dm.org/enketo/enketo-core) [![devDependency Status](https://david-dm.org/enketo/enketo-core/dev-status.svg)](https://david-dm.org/enketo/enketo-core?type=dev) [![Codacy Badge](https://api.codacy.com/project/badge/Grade/dc1c5aaa9267d75cbd2d6714d2b4fa32)](https://www.codacy.com/app/martijnr/enketo-core?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=enketo/enketo-core&amp;utm_campaign=Badge_Grade)
 
 Enketo Core
--- a/package.json
+++ b/package.json
@@ -103,5 +103,8 @@
         "node-forge": "^0.10.0",
         "openrosa-xpath-evaluator": "^2.0.6",
         "signature_pad": "2.3.x"
+    },
+    "volta": {
+        "node": "12.22.1"
     }
 }
--- a/src/widget/range/range-widget.js
+++ b/src/widget/range/range-widget.js
@@ -84,7 +84,7 @@ class RangeWidget extends Widget {
         const html =
             `<div class="widget range-widget">
                 <div class="range-widget__wrap">
-                    <div class="range-widget__current"></div>
+                    <div class="range-widget__current">${this.props.value}</div>
                     <div class="range-widget__bg"></div>
                     <div class="range-widget__ticks"></div>
                     <div class="range-widget__scale">
@@ -183,6 +183,7 @@ class RangeWidget extends Widget {
         props.min = Number( min );
         props.max = Number( max );
         props.step = Number( step );
+        props.value = this.element.value;
         props.vertical = props.appearances.includes( 'vertical' ) || distress;
         props.ticks = !props.appearances.includes( 'no-ticks' );
         props.showScale = distress;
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -103,5 +103,8 @@
         "node-forge": "^0.10.0",
         "openrosa-xpath-evaluator": "^2.0.6",
         "signature_pad": "2.3.x"
+    },
+    "volta": {
+        "node": "12.22.1"
```
MartijnR, line 108: Seems useful. Thanks. Just checking: do you know if Volta could use the information in the `"engines" : { "node": "12", "npm": "6"}` setting instead? That would also help non-Volta users.

README.md
```diff
@@ -1,4 +1,4 @@
-![coverage-shield-badge-1](https://img.shields.io/badge/coverage-74.7%25-yellow.svg)
+![coverage-shield-badge-1](https://img.shields.io/badge/coverage-74.92%25-yellow.svg)
 [![npm version](https://badge.fury.io/js/enketo-core.svg)](http://badge.fury.io/js/enketo-core) [![Build Status](https://travis-ci.com/enketo/enketo-core.svg?branch=master)](https://travis-ci.org/enketo/enketo-core) [![Dependency Status](https://david-dm.org/enketo/enketo-core/status.svg)](https://david-dm.org/enketo/enketo-core) [![devDependency Status](https://david-dm.org/enketo/enketo-core/dev-status.svg)](https://david-dm.org/enketo/enketo-core?type=dev) [![Codacy Badge](https://api.codacy.com/project/badge/Grade/dc1c5aaa9267d75cbd2d6714d2b4fa32)](https://www.codacy.com/app/martijnr/enketo-core?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=enketo/enketo-core&amp;utm_campaign=Badge_Grade)
```
MartijnR, line 2: This is fine, but for the future, I find it easier to leave this out and only update when we do a release to avoid out-of-order PR merge conflicts.

tutorials/20-development.md
```diff
@@ -1,6 +1,6 @@
 ### How to develop Enketo Core
 
-1. install [node](http://nodejs.org/) version 8 until [this issue](https://github.com/enketo/enketo-transformer/issues/53) is closed, and [grunt-cli](http://gruntjs.com/getting-started)
+1. install [node](http://nodejs.org/) version 12, and [grunt-cli](http://gruntjs.com/getting-started)
```
MartijnR, line 3: 👍

src/widget/range/range-widget.js
```diff
@@ -183,6 +183,7 @@ class RangeWidget extends Widget {
         props.min = Number( min );
         props.max = Number( max );
         props.step = Number( step );
+        props.value = this.element.value;
```
MartijnR, line 186: I would call this `loadedValue` to avoid confusion since it's static.

src/widget/range/range-widget.js
```diff
@@ -84,7 +84,7 @@ class RangeWidget extends Widget {
         const html =
             `<div class="widget range-widget">
                 <div class="range-widget__wrap">
-                    <div class="range-widget__current"></div>
+                    <div class="range-widget__current">${this.props.value}</div>
```
MartijnR, line 87: This is a totally fine solution.

It looks like the call to `this.update()` in `_init()` should have loaded the displayed current value as well. That way both the widget value as well as the current value box are set in the same way.

I think the bug appeared in the change listener when adding `if ( this.originalInputValue !== this.value )` to avoid unnecessary change events firing when setting `this.originalInputValue`.

Moving `this.current.textContent = this.value;` outside of that if statement would be an alternative fix to this issue. It avoids having to add a `prop` and sets the display value only in one location.

What do you think?

MartijnR (commented):
Welcome and thank you @eyelidlessness!

package.json
eyelidlessness: This intuitively seemed like something it'd support, but evidently they've [decided](https://github.com/volta-cli/volta/issues/355) against this because the fields have different semantics: `engines` specifies a range, while `volta` (née `toolchain`) pins a specific version.

README.md
eyelidlessness: That makes sense! I had assumed this was expected because it automatically updates on `npm test`. Okay if I add a note in tutorials/20-development.md?

src/widget/range/range-widget.js
eyelidlessness: This makes sense. I iterated on this several times, trying to get a sense of the preferred conventions, but admittedly there's a fair bit to take in, so I decided that getting feedback like this would be more valuable for a first PR. 🙂

src/widget/range/range-widget.js
eyelidlessness: I considered this approach as well. To be honest, I've spent so much time working with component libraries like React that I assumed `props` was the preferred solution for data passed into a component from outside context (albeit implicitly in this case). Which would you prefer?

package.json
MartijnR: Thanks! Let's keep it then.

README.md
MartijnR: Yes, definitely. Thanks!

src/widget/range/range-widget.js
MartijnR: Thanks. Let's do the latter for the consistency with setting default values.

MartijnR:
Thanks!

Issue #579, opened 2018-10-26
Add support for setlocation


MartijnR:
after or during #577 

Documentation: https://github.com/opendatakit/xforms-spec/pull/198

eyelidlessness:
I started on this over the last couple days and have some notes on what I've found so far. Some of this may be obvious/not noteworthy but I'm new to Enketo so I'm still taking a lot in.

- **Affected projects:** `odk:setgeopoint` is essentially a subset of `setvalue` where `value` will populated using the [Geolocation API](https://developer.mozilla.org/en-US/docs/Web/API/Geolocation_API) rather than user input. So I looked through the codebase to understand how `setvalue` is implemented. It turns out that it depends on transformations in [enketo-transformer](https://github.com/enketo/enketo-transformer). I confirmed with @MartijnR that there will need to be similar work there to support the implementation in enketo-core.
- **Transformations:** there are several `setvalue` edge cases handled in enketo-transformer, some of which may also apply to `odk:setgeopoint`, I'm still walking through the relevant test code to determine where they overlap.
- **Testing:**
  - I have an open question about whether there's a preference for parameterizing or repeating test cases where the expected behavior between `setvalue` and `odk:setgeopoint` is identical or nearly so. My hunch is that there will be quite a lot of overlap in enketo-transformer which may benefit from parameterization, which may reduce maintenance burden. And in enketo-core my hunch is that they're more likely to diverge.
  - Regardless of preference on that, to the extent possible, I'm planning to use existing `setvalue` tests and fixtures as a starting point for testing `odk:setgeopoint`. This should help me familiarize myself with the various edge cases and expectations that a solid implementation should consider.
- **Behavior implementation:** Once transformation is handled, the behavior implementation in enketo-core appears to largely be handled in `src/js/calculate.js`. Until transformation is complete, this is pure speculation, but I strongly suspect the work to support `odk:setgeopoint` will be able to perform the relevant Geolocation API lookup—`navigator.geolocation.getCurrentPosition()`, if I'm not mistaken—and defer to the existing `setValue` method from there.
[...]

PR #778, opened 2021-05-08
Implement odk:setgeopoint for odk-instance-first-load events

eyelidlessness:
This implements part of #579, setting a `geopoint` value based on the user's current location on load.

Notes:

- This currently does not work within a `repeat`, discussion in [these](https://github.com/enketo/enketo-core/issues/579#issuecomment-833133293) [comments](https://github.com/enketo/enketo-core/issues/579#issuecomment-833687549).
- This currently depends on a corresponding branch on my fork of `enketo-transformer` until https://github.com/enketo/enketo-transformer/pull/101 is released.
- I generalized the call to `Geolocation#getCurrentPosition`, but it's worth noting that the two calls which already existed had slightly different behavior. I chose the approach that explicitly handled accuracy over the approach which did not. If those were intentionally different I can make those calculations optional.
- I broke up some methods in `calculate.js` to try to reuse as much code from `setvalue` as possible, but if this isn't preferred I'm open to another approach.
- I have chosen a newer distro version in `.travis.yml` in hopes of being able to use the Geolocation API, as the version of Chrome on the previous test run was 30(!) versions behind.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "devDependencies": {
         "cross-env": "^7.0.3",
         "docdash": "^1.2.0",
-        "enketo-transformer": "https://github.com/eyelidlessness/enketo-transformer.git#1617062",
+        "enketo-transformer": "1.43.0",
         "eslint-plugin-jsdoc": "^32.3.0",
         "grunt": "^1.3.0",
         "grunt-concurrent": "^3.0.0",
--- a/package.json
+++ b/package.json
@@ -50,7 +50,7 @@
     "devDependencies": {
         "cross-env": "^7.0.3",
         "docdash": "^1.2.0",
-        "enketo-transformer": "1.42.0",
+        "enketo-transformer": "https://github.com/eyelidlessness/enketo-transformer.git#1617062",
         "eslint-plugin-jsdoc": "^32.3.0",
         "grunt": "^1.3.0",
         "grunt-concurrent": "^3.0.0",
--- a/src/js/calculate.js
+++ b/src/js/calculate.js
@@ -246,6 +246,7 @@ export default {
 
         const newModelValue = props.dataNodesObj.getVal();
 
+        // This is okay for an xforms-value-changed action (may be no form control)
         if ( !control ) {
             return;
         }
@@ -264,9 +265,6 @@ export default {
             if ( control.type !== 'hidden' && config.validateContinuously === true ) {
                 this.form.validateInput( control );
             }
-        } else {
-            // This is okay for an xforms-value-changed action (may be no form control)
-            // console.log( 'no form control found' );
         }
     },
 
--- a/src/js/calculate.js
+++ b/src/js/calculate.js
@@ -64,7 +64,7 @@ export default {
                 } else if ( control.type === 'hidden' ) {
                     /*
                      * This case is the consequence of the  decision to place calculated items without a visible form control,
-                     * as a separate group (.or-calculated-items, or .or-setvalue-items), instead of in the Form DOM in the locations .
+                     * as a separate group (.or-calculated-items, or .or-setvalue-items, or .or-setgeopoint-items), instead of in the Form DOM in the locations .
                      * This occurs when update is called with empty updated object and multiple repeats are present.
                      */
                     dataNodes.forEach( ( el, index ) => {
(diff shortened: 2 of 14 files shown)


package.json
```diff
@@ -50,7 +50,7 @@
     "devDependencies": {
         "cross-env": "^7.0.3",
         "docdash": "^1.2.0",
-        "enketo-transformer": "1.42.0",
+        "enketo-transformer": "https://github.com/eyelidlessness/enketo-transformer.git#1617062",
```
lognaturel, line 53: As far as I can tell from doing a quick `blame`, @MartijnR has always done a Transformer release before merging something to dependent repos. Now that we're looking at a larger and less coordinated set of devs who don't all have permissions to do releases, I think we should pick a process and document it. E.g. Keep PRs like this one in draft state until all Transformer PRs have been reviewed and merged, request a Transformer release from @MartijnR, then update the Transformer version and mark PR as ready.

src/js/calculate.js
```diff
@@ -90,46 +91,53 @@ export default {
...
-        let nodes = [];
-
+    _getNodesForAction( event, action ) {
         if ( event.type === new events.InstanceFirstLoad().type ) {
             // We ignore relevance for the data-instance-first-load, as that will likely never be what users want for a default value.
             // Do not use getRelatedNodes here, because the obtaining (and caching) of nodes inside repeats is (and should be) disabled at the
             // time this event fires.
             //
-            // We change the order by first evaluating the non-formcontrol setvalue directives (in document order), and then
+            // We change the order by first evaluating the non-formcontrol setvalue/setgeopoint directives (in document order), and then
```
lognaturel, line 103: "setvalue/setgeopoint directives" -> actions? Similar to what we discussed on at https://github.com/enketo/enketo-transformer/pull/101#discussion_r627570573

src/js/calculate.js
```diff
@@ -90,46 +91,53 @@ export default {
...
         } else if ( event.type === new events.XFormsValueChanged().type ) {
             const question = event.target.closest( '.question' );
-            nodes = question ? [ ...question.querySelectorAll( `[data-setvalue][data-event*="${event.type}"]` ) ] : nodes;
+
+            return question ? [ ...question.querySelectorAll( `[data-${action}][data-event*="${event.type}"]` ) ] : [];
+        }
+    },
+
+    /**
+     * Runs <setvalue> actions.
```
lognaturel, line 125: "<setvalue> actions" -> "actions"

src/js/calculate.js
```diff
@@ -90,46 +91,53 @@ export default {
...
+     *
+     * @param {'setvalue' | 'setgeopoint'} type - the nodes relevant to the action.
```
lognaturel, line 127: "the nodes relevant to the action." Don't think so?

src/js/calculate.js
```diff
@@ -90,46 +91,53 @@ export default {
...
+     * @param {CustomEvent} [event] - the event type that triggered the setvalue action.
+     */
+    performAction( type, event ) {
+        if ( !event ) {
+            return;
         }
 
+        if ( !this.form ) {
+            throw new Error( 'Setvalue module not correctly instantiated with form property.' );
```
lognaturel, line 136: "Setvalue module" -> "Action"? I'm not really clear on what "module" refers to here.

src/js/calculate.js
```diff
@@ -208,10 +243,18 @@ export default {
 
         const newModelValue = props.dataNodesObj.getVal();
 
+        // It's possible there's a better way to do this, but this works around
+        // an error where `odk:setgeopoint` inside a model can't find an associated
+        // `.question` in the DOM.
+        if ( !control || this.form.input.getWrapNode( control ) == null ) {
```
lognaturel, line 249: I can't find it now but I feel like I remember you mentioning you thought there could be a deeper bug or error state that needs handing here even for `setvalue`. Does that sound right?

src/js/form.js
```diff
@@ -1179,6 +1188,6 @@ Form.prototype.goToTarget = function( target ) {
  * @type {string}
  * @default
  */
-Form.requiredTransformerVersion = '1.42.0';
+Form.requiredTransformerVersion = 'https://github.com/eyelidlessness/enketo-transformer.git#1617062';
```
lognaturel, line 1191: @Martijnr I think you mentioned this as extra protection against a Transformer version mismatch? How often is it the case that you bump the dependency without also bumping this? Has it caught mistakes before?

I can't find this documented beyond [just being listed](https://enketo.github.io/enketo-core/Form.html#.requiredTransformerVersion). If it is helpful, I think it would be a good thing to add to the readme or something.

package.json
eyelidlessness: This makes sense! I've converted this and #779 to draft PRs. <small>(Could they make that link-button any more invisible?)</small>

src/js/calculate.js
eyelidlessness: I think you may be thinking of the `repeat` async issue, but looking at this again actually helped me discover that I'd missed a corresponding change in `input.js` that makes this check unnecessary. Thanks!

src/js/form.js
eyelidlessness: I found this check a bit odd, I don't often see version checks like this because versions are pinned in `package.json`. I understand (especially from working on this task) that `enketo-core` is tightly coupled to `enketo-transformer` and depends on certain functionality, but I wonder if there might be a way to express that more explicitly, e.g. by checking for a certain set of explicit, named features that the transformer supports? Obviously out of scope for this PR, but may also help reduce churn here.

src/js/calculate.js
```diff
@@ -222,7 +265,7 @@ export default {
                 this.form.validateInput( control );
             }
         } else {
-            // This is okay for a setvalue/xforms-value-changed directive (may be no form control)
+            // This is okay for an xforms-value-changed action (may be no form control)
```
MartijnR, line 268: Just in case this is a useful comment in the future, perhaps move to the if(!control) block.

This `else{}` can then be removed, I think

src/js/calculate.js
MartijnR: Module is the whole file (an ES module). It's a weird check perhaps and I think there is some historic reason for it, probably related to testing. I won't work like a regular ES module unless it's added like this: https://github.com/enketo/enketo-core/blob/c35aaba086876ff0b2588270f2fad5ed71f16c25/src/js/form.js#L199

May be worth revisiting the Form.addModule approach in the future.

src/js/form.js
lognaturel: Yikes.

lognaturel:
Can now be updated to use [transformer 1.43](https://github.com/enketo/enketo-transformer/releases/tag/1.43.0)! 🎉

MartijnR:
Thanks!! 👍

PR #779, opened 2021-05-08
Implement odk:setgeopoint for xforms-value-changed events

eyelidlessness:
Branched from #778 ([see diff](https://github.com/eyelidlessness/enketo-core/compare/feature/setgeopoint...eyelidlessness:feature/setgeopoint-xforms-value-changed)), also part of #579.

This only consists of test changes, as the implementation to support `odk-instance-first-load` also satisfied this case.


lognaturel:
I personally really like this separation. 👍 Will be really easy to review once #778 is merged.

eyelidlessness:
> I personally really like this separation. 👍 Will be really easy to review once #778 is merged.

Oh I'm glad 🙂  Quite a few times I've been in a situation where I've had to chain multiple PRs to accomplish some larger set of interdependent tasks, and this approach (open the PR with the previous branch's changes, but link to a diff showing just the current changes) has been the best I've come up with.

lognaturel:
Needs a little conflict resolution!

I personally would have tried to either rebase on `master` or start back from `master` and cherry pick commits specific to this branch back onto it after #778 was merged. But I understand that can also be pretty fiddly and if we're going to squash on merge anyway it's not a big deal.

As a general note, I do typically review commit-by-commit and find a lot of value in looking at smaller chunks (when possible). Mentioning that so it's clear there can be value to smaller individual commits (whereas there wouldn't be if all reviewers just looked at the final diff).

lognaturel (approved):

eyelidlessness:
@lognaturel I appreciate you preferring commit by commit! I've always preferred to keep the messier log than squash commits. Clearing up these conflicts was more messy than I anticipated mostly because I forgot I had another branch laying around for reference... and partly because I missed an additional conflict cleaning that up. Extra fun is... [there wasn't actually any conflict](https://github.com/eyelidlessness/enketo-core/compare/dbd991b...eyelidlessness:feature/setgeopoint-xforms-value-changed), but getting sloppy with my branches must have caused that.

lognaturel:
We'll just extra squash this one. 😅

MartijnR (approved):
Great! Thank you! :shipit:

PR #782, opened 2021-05-20
Simplify test and getIndex fix for triply nested case

lognaturel:
I'm late to your #723 party, @MartijnR and @FrankApiyo but did want to take a moment to understand what was going on since I know repeats are complex.

This PR is mostly meant to check my understanding. I think the simplifications proposed might also help us in the future if/when this code needs to be revisited but will be interested to hear whether others agree.

I also moved the test. In the repeats tests, there's a `cloning` test block that this new test was previously in but then there are many tests that appear related to cloning outside of that block, too. @MartijnR is there meaning to that organization or is it just that the block was forgotten about at some point? I moved the new test to after `'In an empty form it creates the first repeat instance automatically` because there are other tests around there about nested repeats and the issue turned out to be about missing repeat instances on outer repeat creation.

Some additional clarifications:
- My understanding of the bug is that because the info nodes were not being counted correctly (the count was increasing even if there wasn't an info node there), there were nodes that were supposed to be cloned that weren't. Is that a reasonable one sentence explanation? If not, what would one be?
- `getIndex` computes an absolute index of repeat instances or repeat info of a particular name across the entire primary instance? What's the high-level reason for it? E.g. is it used to index into linear collections that keeps track of all instances of one repeat regardless of their nesting structure?
- In the conversation around this issue, @MartijnR you mentioned some kind of recursive algorithm for `getIndex`. I don't see any recursion there, am I missing it? It seems it walks the node tree iteratively. The inner loop counts repeat instance that are siblings of the node that was passed in and the outer loop does the same count in any outer repeats that may exist. Did I get that roughly correctly?
- I'm having trouble understanding the initialization of the count in `getIndex`.  I would expect that initializing using `repeat-number` and then counting `previousElementSibling`s would double count but clearly that's not the case. What am I missing?
- @MartijnR mentioned that there might be value in splitting `getIndex`. This feels related to some questions we've run into around `set`/`update` for the records queue and `saveRecord`/`submitRecord` for online and offline views. My sense is that it's dangerous to have two functions that have largely duplicated code with subtle differences. It requires a lot of concentration to consistently make updates in both in parallel and it can be difficult for the call site to make the right choice. I've proposed alternate simplifications that hopefully increase readability without needing two functions.
- Doesn't `getInfoIndex` do exactly the split you describe? Except I think it wouldn't work in nested cases? I feel it illustrates what I described above -- it has a subtly different signature (can return null), and clearly wasn't used everywhere it could have been. I can confirm that using its approach in `getIndex` would lead to all tests passing except those having to do with nested repeats. I propose removing the duplication.


src/js/repeat.js
```diff
@@ -133,40 +133,23 @@ export default {
             return 0;
         }
 
-        let checkEl = el.parentElement.closest( '.or-repeat' );
```
lognaturel, line 136: There's currently no test coverage for a case where the `closest` call would return something other than the element itself. As far as I can tell, the function should always be given a repeat or repeat info so that would never happen. I could be missing something, though, and this seems it could be risky.

src/js/repeat.js
```diff
@@ -133,40 +133,23 @@ export default {
...
-        const info = el.classList.contains( 'or-repeat-info' );
-        let count = info ? 1 : Number( el.querySelector( '.repeat-number' ).textContent );
-        const name = el.dataset.name || el.getAttribute( 'name' );
+        const isInfoElement = el.classList.contains( 'or-repeat-info' );
 
+        const toCount = isInfoElement ? `.or-repeat-info[data-name="${el.dataset.name}"]` : `.or-repeat[name="${el.getAttribute( 'name' )}"]`;
```
lognaturel, line 138: The big win for me is centralizing all differences based on whether we're dealing with an info element or a real repeat instance. This way the algorithmic part is entirely shared and that reduces the cognitive load a lot for me.

I also really got tripped up by the `name` declaration before. The fact that it was picking up a different name based on whether or not the element was an info element was implicitly based on the presence of `el.dataset.name`. That felt like it added a lot of mental hoops to jump through.

eyelidlessness: 1. It wasn't entirely clear to me what a "repeat info" is supposed to mean. After some hunting, I found this in the docs:

    > Each repeat series has a sibling .or-repeat-info element that stores info that is relevant to that series.

    This still didn't clear much up for me, so I went and looked for these nodes as rendered. To validate what I think I understand, I think they provide two purposes?

    - Render a button bound to create new repeat instances
    - Specify a model reference (`data-name`) to the associated repeat structure, which is repeated when the button is clicked

2. It's not clear to me why that element would factor into a count. I would think this condition is only to determine the repeat's name, but the selector would be looking only for `.or-repeat` elements with the specified name, e.g.

    ```js
    const toCount = isInfoElement ? `.or-repeat[name="${el.dataset.name}"]` : `.or-repeat[name="${el.getAttribute( 'name' )}"]`;
    ```

    But it's entirely possible I'm misunderstanding what is being counted.

3. Maybe overly nitpicky, but I think it would be easier to follow the rest of this if the name specified that it's a selector. Names like `toCount` confuse me because I would expect them to specify some sort of iterable structure (set, list, queue, etc) of data to be counted.

src/js/repeat.js
```diff
@@ -133,40 +133,23 @@ export default {
...
+        let count = isInfoElement ? 1 : Number( el.querySelector( '.repeat-number' ).textContent );
```
eyelidlessness, line 139: (Not asking for a change here, just making a note.) I was surprised to see a method named `getIndex` which appears to be 1-based (in JavaScript). I see now that [the return value](https://github.com/enketo/enketo-core/pull/782/files#diff-ca036963ec5e89fdaa73c05b2ff5381f3f22d89c8ca700daa74aee65e0821dc8R151) then subtracts to make it 0-based. Also out of scope, but this is another case where I think mixing state and DOM makes code hard to reason about. Of course the user-facing numbers would be 1-based, but that's a representation of state, not *actual* state.

eyelidlessness (changes requested):
I added a couple minor change requests, some clarifying questions, and a few general notes that are out of scope for this PR. In the same vein, I'll generally reiterate my preference to decouple state and its representation in the DOM. Trying to determine state _from_ the DOM is error prone and subject to all sorts of unexpected breakage even with good test coverage. Whereas, with decoupled state this method would simply be a call like `state.items.indexOf(item)`.

src/js/repeat.js
lognaturel: 1. I also had to go hunting for that same doc and came to a similar conclusion. Would be good to get confirmation that this covers its full purpose and think about augmenting the existing docs a bit for posterity.

2. My understanding is that counting the `repeat-info` elements is useful for knowing how many repeat sequences of a particular type there are. Let's say you have a `butterfly` repeat nested in a `meadow` repeat. `meadow 1` has 10 butterflies, `meadow 2` has 4 butterflies. I believe there will be a total of 14 butterfly repeat instances and 2 corresponding `repeat-info`/buttons for adding butterflies/`butterfly` sequences. Basically it's the count of the immediate parent repeat, I think. I'm not sure, though, and it's one of the questions I've asked.

As far as I understand, what you wrote wouldn't make sense because a repeat element wouldn't have a `dataset.name`.

3. That's a good point. Does `toCountSelector` help?

test/spec/repeat.spec.js
```diff
@@ -182,6 +162,22 @@ describe( 'repeat functionality', () => {
         expect( form.view.html.querySelectorAll( a )[ 0 ].querySelectorAll( b ).length ).toEqual( 1 );
     } );
 
+    // https://github.com/enketo/enketo-core/issues/720
+    it( 'adds nested repeats up to three deep correctly', () => {
```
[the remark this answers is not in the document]
lognaturel: Haha, whoops, I do agree with you. Will change once we agree on other changes.

src/js/repeat.js
eyelidlessness: 1. I agree, it would be good to add clarity to the docs. I think it would also be helpful to add a comment describing its purpose in `enketo-transformer`, as that's where I looked first.

2. I realized after posting my initial review that I may have been mistaken here. I think my confusion mostly arises from naming, and what this method is actually doing. It does make sense that one might want to get an "index" of the "info" button (for certain purposes), but it doesn't make sense (to me) that one would want to "count" it.

3. Definitely better, notwithstanding the general count/index naming confusion.

- - -

Regarding point 2, I tried to step through the test/implementation to understand what's going on, and I have to admit to only being more confused than when I started reviewing. It's not clear to me what `getIndex` is actually intended to "count", or what an "absolute index" (per JSDoc) represents.

I added some logging on calls from `add`, to hopefully gain some more insight. Here's what I see running the "three deep" test:

  <details>
    <summary>with current logic</summary>
<pre><code>
LOG: 'setup'
LOG: Object{name: '/data/outer', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer', isInfoElement: false, result: 0}
LOG: Object{name: '/data/outer/inner', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer/inner', isInfoElement: false, result: 0}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: false, result: 0}
LOG: Object{name: '/data/outer/inner', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer/inner', isInfoElement: false, result: 0}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: false, result: 0}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: false, result: 0}
LOG: '/setup'
LOG: 'click third'
LOG: Object{name: '/data/outer/inner/third', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: false, result: 1}
LOG: '/click third'
LOG: 'click inner'
LOG: Object{name: '/data/outer/inner', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer/inner', isInfoElement: false, result: 1}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: true, result: 1}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: false, result: 2}
LOG: '/click inner'
LOG: 'click outer'
LOG: Object{name: '/data/outer', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer', isInfoElement: false, result: 1}
LOG: Object{name: '/data/outer/inner', isInfoElement: true, result: 1}
LOG: Object{name: '/data/outer/inner', isInfoElement: false, result: 2}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: true, result: 2}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: false, result: 3}
LOG: '/click outer'
</code></pre>
  </details>

  Here's what it looks like if I add two more clicks on `third`, once immediately and again after the `outer` click:

  <details>
    <summary>with added clicks</summary>
<pre><code>
LOG: 'click third'
LOG: Object{name: '/data/outer/inner/third', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: false, result: 1}
LOG: '/click third'
LOG: 'click third again'
LOG: Object{name: '/data/outer/inner/third', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: false, result: 2}
LOG: '/click third again'
LOG: 'click inner'
LOG: Object{name: '/data/outer/inner', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer/inner', isInfoElement: false, result: 1}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: true, result: 1}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: false, result: 3}
LOG: '/click inner'
LOG: 'click outer'
LOG: Object{name: '/data/outer', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer', isInfoElement: false, result: 1}
LOG: Object{name: '/data/outer/inner', isInfoElement: true, result: 1}
LOG: Object{name: '/data/outer/inner', isInfoElement: false, result: 2}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: true, result: 2}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: false, result: 4}
LOG: '/click outer'
LOG: 'click third after outer'
LOG: Object{name: '/data/outer/inner/third', isInfoElement: true, result: 0}
LOG: Object{name: '/data/outer/inner/third', isInfoElement: false, result: 3}
LOG: '/click third after outer'
</code></pre>
  </details>

These values are confusing, I've tried/given up half a dozen times to explain what I think they might mean. At this point I think I'm better off just sharing my confusion and hope for clarity in replies.

eyelidlessness: I keep coming back to try to understand this and I think I have it, but if I do I think this may be a candidate for refactoring sooner rather than later. (Typing on phone now because I had to step away from desk, apologies if it’s a mess.)

I’d need to add a more complex test/model to validate all of this, but:

- It seems for “info” elements the index likely amounts to depth within a given clone action (or depth times however many repeats of the same path are specified? Are adjacent/redundant repeats of the same path supported?)
- For repeated model values it seems to correspond to position in the record top-to-bottom regardless of depth/structure. Which is to say I would think the final click in my modification pushed the inner/third and outer/inner/third indices to 5 and 6 respectively.

If I’m understanding this correctly (and I very well may still not!), I think this is two functions which just happen to produce mostly expected output by luck. And if that’s the case, I’d favor making this less DRY just for the opportunity to name and document the different semantics.

More than that, I think I would recommend starting to decouple state/DOM more aggressively than I would have at first glance. It doesn’t necessarily require adopting purpose built solutions right away (though it will probably add weight to those options if/when there’s bandwidth to discuss them). Just keeping a data structure of current state that’s represented in the DOM without referencing it, and clearly delineating how data flows.

It could also be a good opportunity to investigate framework-agnostic state management solutions (I have several in mind), but I’m hesitant to push for that just now for a bunch of reasons, unless there’s an appetite for it.

Ultimately the initial goals I have in mind, in order of hierarchy of needs:

- Clarify what this is actually expected to do, in discussion
- Prioritize naming/documenting improvements that reflect that clarification
- Identify and plan steps to allow reasoning with state separately from rendering
- Consider tools that might aid state/view in the current stack
- Plan longer term to evaluate whether tools built for these purposes might be introduced gradually to begin establishing a pattern for future iteration/maintenance

src/js/repeat.js
```diff
@@ -133,40 +133,23 @@ export default {
...
-        if ( !this.form.repeatsPresent ) {
-            return 0;
-        }
-        if ( !repeatInfo || !repeatInfo.classList.contains( 'or-repeat-info' ) ) {
-            return null;
-        }
-        const name = repeatInfo.dataset.name;
-
-        return [ ...repeatInfo.closest( 'form.or' ).querySelectorAll( `.or-repeat-info[data-name="${name}"]` ) ].indexOf( repeatInfo );
-    },
```
MartijnR, line 169: O M G.... this existed??? Nice to have reviewers! 👍

src/js/repeat.js
MartijnR: Thanks! This is great. I agree that toCountSelector would be even clearer.

MartijnR: The repeat-info was introduced when we started supporting zero repeats (but that's all I remember).

MartijnR (approved):

MartijnR:
> or is it just that the block was forgotten about at some point?
Yes
> I don't see any recursion there, am I missing it? It seems it walks the node tree iteratively. 
Yes, indeed. I used the wrong word to describe that the code should work for any level of nesting (to explain that the workaround proposed in the previous PR seemed not right

Thanks for this PR and reviewing!

src/js/repeat.js
lognaturel: It really is. Sorry you had to wade through so much of this on your own! It's a lot.

src/js/repeat.js
lognaturel: Thanks, @MartijnR, that's helpful.

@eyelidlessness and I went over the logs he pulled again looking at the UI and we are feeling like we can now consolidate that with our understanding of `getIndex`. The confusion came from the last added click to add an instance of the third level of repeats. It adds a `third` instance to the first instance of `inner`, not to the last one.

lognaturel:
Thanks so much for taking a look, @MartijnR, and for putting up with my wall of questions. My spidey-sense tells me we'll be back in this code so I think it's a good investment to get context now.

I feel like I'm at least confident I understand what `getIndex` does and how it works. I don't have a sense of why this index is used (as opposed to e.g. an index within a parent). If you can give us a high-level sense or a hint on where to look, @MartijnR, I think that'd good to know. But also not critical at the moment. We have a lot more context and breadcrumbs now.

In response to @eyelidlessness's comment about count vs. index, I've tried changing the accumulator's name to `predecessorCount` and I find that helpful. That also led me to move the adjustment to go from a 1-based index to a 0-based index up in the initialization which I think makes more sense. That change helped me understand why the initialization works: the inner while loop doesn't add to the count in the case of a non-nested repeat. The loop condition is checking for repeat siblings but then the `querySelectorAll` call looks for *children* of the element being checked that match the original element's name. If I got all that right, the one unfortunate thing is that there is iteration over every instance of a repeat that adds 0 every time even if it's not nested. I'm not immediately coming up with a way to remove that and it doesn't feel like a big deal.

eyelidlessness (approved):

eyelidlessness:
This whole process has been extraordinarily clarifying. Thank you @lognaturel both for revisiting it and helping me get a better understanding of what’s happening here. The only thing I want to add for posterity, which was additionally clarifying in our check in today, is that the thing that helped us understand what happens in this bit of code is the “absolute” index is an index of all repeated/nested references as if they had been flattened into a single list, and that that term flatten helped make it clear.

MartijnR:
>‪ I don't have a sense of why this index is used (as opposed to e.g. an index within a parent). If you can give us a high-level sense or a hint on where to look, @MartijnR, I think that'd good to know. But also not critical at the moment. We have a lot more context and breadcrumbs now.

This was a very early choice based on easy querySelectorAll (or jQuery) searches, as opposed to determining indices at every nested step. Not sure if that still makes sense, but it was never revisited by me.

MartijnR:
Thanks so much to you both. This looks like a big improvement! 🎉

lognaturel:
> This was a very early choice based on easy querySelectorAll (or jQuery) searches

Ah, got it. Makes perfect sense, thank you!

@eyelidlessness it might also help to consider that repeats in general but especially nested repeats are not a big priority for web form users. Nested repeats in particularly generally imply an enumerator-mediated context and even then they're relatively rare. Most of the time when forms with nested repeats end up in Enketo they were authored for Collect and are being used for edits.

I still wouldn't be surprised if we did work in this area sooner than later but major refactoring might not be so compelling unless it benefits no-repeat or single-repeat cases significantly.

PR #783, opened 2021-05-20
Use new data-form-id attribute for formId

### LINKED ISSUE #781: Update transformer to v1.43 and change access of form's id to use new data-form-id attribute
Once https://github.com/enketo/enketo-transformer/pull/105 is merged and released.

https://github.com/enketo/enketo-core/blob/master/src/js/form.js#L160

lognaturel:
Closes #781

See https://github.com/enketo/enketo-transformer/issues/100

### DIFF
--- a/src/js/form.js
+++ b/src/js/form.js
@@ -158,7 +158,7 @@ Form.prototype = {
      * @type {string}
      */
     get id() {
-        return this.view.html.id;
+        return this.view.html.dataset.formId;
     },
     /**
      * To facilitate forks that support multiple constraints per question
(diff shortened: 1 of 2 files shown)


eyelidlessness (approved):

src/js/form.js
```diff
@@ -158,7 +158,7 @@ Form.prototype = {
      * @type {string}
      */
     get id() {
-        return this.view.html.id;
+        return this.view.html.attr( 'data-form-id' );
```
eyelidlessness, line 161: This should probably be `this.view.html.dataset.formId`

eyelidlessness (changes requested):

src/js/form.js
eyelidlessness: And this should have a test

lognaturel: Yiiiikes! I am really not to be trusted. 🙏🙏🙏

eyelidlessness (approved):

MartijnR (approved):
Thanks!

Issue #484, opened 2017-12-05
Add support for locales in number inputs?


MartijnR:
https://www.ctrl.blog/entry/html5-input-number-localization

MartijnR:
I don't really want to move away from `<input type="number">` to `<input type="text"/>` but may be the mask can be tweaked depending on the decimal separator used and supported in native number inputs.

It looks like thousand separators aren't really supported.

MartijnR:
Although it is actually possible to show the numeric keyboard on mobile devices for text input.... (same as appearance="numbers")

eyelidlessness:
As mentioned in #792, I think this should be pretty simple to address now, using `Intl.NumberFormat`.

PR #790, opened 2021-06-15
fixed: OC: Text Multiline item with long values loads with widget condensed

theywa:
[#484](https://github.com/OpenClinica/enketo-express-oc/issues/484)

### DIFF
--- a/src/widget/textarea/textarea.js
+++ b/src/widget/textarea/textarea.js
@@ -1,5 +1,7 @@
 import Widget from '../../js/widget';
 import events from '../../js/event';
+import { MutationsTracker } from '../../js/dom-utils';
+
 
 /**
  * Auto-resizes textarea elements.
@@ -23,7 +25,11 @@ class TextareaWidget extends Widget {
                 this._resize( el );
             }
         } );
-        textareas.forEach( this._resize.bind( this ) );
+        // workaround for resize issue on first time load textarea with existing value on enketo-express
+        new MutationsTracker( this.element ).waitForQuietness()
+        .then( () => { 
+            textareas.forEach( this._resize.bind( this ) );
+        } );
         this.element.addEventListener( events.PageFlip().type, event => {
             const els = event.target.querySelectorAll( 'textarea' );
             els.forEach( this._resize.bind( this ) );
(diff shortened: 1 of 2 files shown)


src/widget/textarea/textarea.js
```diff
@@ -23,7 +25,16 @@ class TextareaWidget extends Widget {
...
+        // workaround for resize issue on first time load textarea with existing value on enketo-express
+        const mutationsTracker = new MutationsTracker(),
+        form = document.querySelector( 'form' ),
+        cls = 'first-load-textarea',
+        classChange = mutationsTracker.waitForClassChange( form, cls );
+        form.classList.add( cls );
+        classChange.then( () => {
+            textareas.forEach( this._resize.bind( this ) );
+            form.classList.remove( cls );
+        } );
```
MartijnR, line 37: Would not adding the class, and instead using something like this work as well?:

```js
const mutationsTracker = new MutationsTracker(); 
mutationsTracker.waitForQuietness()
    .then( () => { .... }
```

If it does it may be worth checking if passing `this.element` to the constructor would work as well. This would help us in understanding the issue better (and I think using that would be better).

theywa: yes, you're right it works. I just update the code also passing the `this.element` to the constructor. So this `.waitForQuietness` method will be fired after no more change happen to the el set on the contructor based on the list of mutation set on the MutationObserver, right?

CMIIW just want to share, I think there is a loophole in the logic in the `checkInterval` on MutationsTracker contructor
``const checkInterval = setInterval( () => {
            if ( previousMutations === mutations ){
                this.quiet = true;
                mutationObserver.disconnect();
                clearInterval( checkInterval );
            } else {
                this.quiet = false;
                previousMutations = mutations;
            }
        }, 100 );
``
from what I debug when try to learn it, it looks like the else condition never fired since the increment set on creation MutationObserver is not working(duplicate variable name).

MartijnR (commented):
Sorry, I reviewed this 10 days ago but forgot to submit!

theywa:
@MartijnR it's okay, sorry for the late response. I just add the fix for duplicate variable names.

MartijnR (approved):
Thanks!

PR #791, opened 2021-06-15
Convert tests from Jasmine to Mocha/Chai/Sinon

eyelidlessness:
This change makes testing in enketo-core more consistent with testing in centro, enketo-transformer and enketo-express (and likely other Enketo projects, those are just the ones I'm most familiar with).

### Benefits/motivation

This should be a development experience improvement, bringing API/test framework feature parity and reducing the cognitive load of switching between projects.

### Risks

Quite a lot of the necessary changes were identified by project-wide search for relevant interface changes (e.g. `toEqual` -> `.to.equal`). However some of those changes were not 100% API-equivalent. For instance some changes of `toEqual` needed to be changed to `to.deep.equal` when comparing reference types.

While those changes were easy to identify (those tests didn't pass with the naive `.to.equal` replacement), the inverse (`.not.toEqual` -> `.not.to.deep.equal`) is not as easily identified, and may result in tests passing where they should not.

I've done my best with the tools available to manually verify that I addressed these cases, but it's certainly possible that there are some I've missed.

It's possible that there are other API inconsistencies of this sort lurking, though feels unlikely based on my reading of Jasmine and Chai docs.


test/karma.conf.js
```diff
@@ -146,7 +144,5 @@ module.exports = config => {
         // Continuous Integration mode
         // if true, Karma captures browsers, runs the tests and exits
         singleRun: false,
-
-        browserify: {},
```
eyelidlessness, line 150: Note: these changes weren't strictly necessary, but bring the config more in line with enketo-express.

lognaturel:
@matthew-white See, you're not alone in the land of migrating test libraries. 😅 Tagging you mostly as an FYI that we're trying to narrow down the testing tools used across Enketo. Naturally, having some alignment with Central would also be great. Here assertions are done with `chai` which seems pragmatic given its adoption and maintenance status. It deviates from Central's use of `should` but that doesn't seem like a big barrier. Anything else you think we should consider?

eyelidlessness:
Oh is there more test framework migration afoot? I’m happy to help if needed, since my head is very much in this game at the moment.

lognaturel:
I didn't think about it until now because it's not identical but @matthew-white is going from https://github.com/eddyerburgh/avoriaz to https://github.com/vuejs/vue-test-utils because of Vue ecosystem changes.

src/js/date.js
```diff
@@ -0,0 +1,69 @@
...
+ * @see {@link https://github.com/enketo/enketo-xpathjs/blob/e6f6d3d1395e41066b7d0c142a8b5fbb5a0e8098/src/date-extensions.js}
+ *
+ * Note: this module was introduced to address this comment in
+ * `enketo/openrosa-xpath-evaluator`:
+ *
+ * ```js
+ * // TODO probably shouldn't be changing Date.prototype
+ * ```
+ *
+ * The logic was copied from `enketo/enketo-xpathjs` as that appeared to be the
```
lognaturel, line 14: As a historical note, `enketo-xpathjs` was the original Enketo XPath evaluator that was in use until earlier this year. It was a JS evaluator with a few extensions including this date one. It was pretty slow. Now `openrosa-xpath-evaluator` uses browsers' evaluators which is much faster. I would expect to see this behavior in the evaluator and am surprised to see it copied here. Did this need to happen because the monkeypatching didn't play nice with the new test libs? Would the idea be to bring this back to `openrosa-xpath-evaluator`?

lognaturel:
Looks great! One inline question about the date-related change.

src/js/date.js
eyelidlessness: We discussed this in our checkin, but for posterity: the issue was that the monkey-patching was happening implicitly in prior test runs given Jasmine's deterministic test order. We decided that this duplication is a risk, and instead I'll directly import the relevant monkey-patch where needed (which we'll also switch to plain functions later).

lognaturel (approved):
Looks great! Thanks for making the date change. We've already discussed with @MartijnR and I don't think there's anything controversial here but we can let a day go by to see if anyone else wants to chime in and then merge. We don't want to sit on this one too long because there are lots of opportunities for conflicts.

test/spec/date.spec.js
```diff
@@ -1,64 +0,0 @@
-/**
- * @module date.spec
- * @see {@link https://github.com/medic/cht-core/blob/2c1db1618bdafc5ec14c3a27aa4a37249fcc1b4a/webapp/tests/mocha/unit/enketo/medic-xpath-extensions.spec.js}
- *
- * Note: These tests were copied from `medic/cht-core` as they appeared to be
- * the only relevant existing tests of this functionality. They've been adapted
- * to:
```
lognaturel: Are these tests in the evaluator? If not, would be good to add them.

eyelidlessness: Went one further and got us on track to stop relying on the prototype monkeypatch! https://github.com/enketo/openrosa-xpath-evaluator/pull/126

Issue #486, opened 2017-12-21
Cannot GET /transform


KishanIos007:
i am trying to use ``enketo-core`` in local server.. i follow the following steps:

```
1) clone the repository 
2) cd to the enketo-core-master
3) npm intsall
4) grunt --v
5) npm start

```

after doing this i am still not able to create the css and js in build folder.after some r & d i fount that it is running on port 8005 and in request url it is port number 8085 so i change the port number in app.js and again repeat the steps from 3 to 5 now css and js is created in build folder but when i click on the first url (see the image)

![capture](https://user-images.githubusercontent.com/18045135/34242015-50c54ea0-e63f-11e7-87e1-2ea6f3023514.PNG)

i get the following error in developer console (see the image)

![1](https://user-images.githubusercontent.com/18045135/34242046-7cb36bc8-e63f-11e7-9ede-59e9a1b13494.PNG)

see this also u will get better idea

![2](https://user-images.githubusercontent.com/18045135/34242094-ae29f99c-e63f-11e7-84b5-c320f87ba09f.PNG)

**UPDATE**

I remove the ``/transform`` from the url but still nothing is showing 😢 

Please can anyone help me out ?

Thanks in advance.

KishanIos007:
@cloudchen  can u please help me

KishanIos007:
@MartijnR  can u please help me

alxndrsn:
@KishanIos007 You need to be running https://github.com/enketo/enketo-transformer as well

alxndrsn:
@KishanIos007 if you clone the repo at https://github.com/enketo/enketo-transformer, and then start that server as well as enketo-core, then the demo page should work for you.

MartijnR:
enketo-transformer has actually been added as a dev dependency so if `npm install` succeeded, you could use `grunt develop` instead of `npm start` to get it all working together without having to clone enketo-transformer. 

See detailed steps here: https://github.com/enketo/enketo-core#how-to-develop-enketo-core

alxndrsn:
@MartijnR thanks for clarifying - very neat!  Can we add enketo-transformer to `npm start`?  Starting only `enketo-core` doesn't seem very useful.

KishanIos007:
can i use enketo-core as node_module in ionic ?? 
[...]

PR #794, opened 2021-06-23
fixed: Opening to item not working for some forms

theywa:
[#486](https://github.com/OpenClinica/enketo-express-oc/issues/486)


theywa:
Hi @MartijnR
After debugging it, it looks like there is a different HTML structure for simple form(without section and fieldset) and complex form with pages or pages-form theme. This issue happens in the simple form where the .closest method on flipToPageContaining (page.js) failed to get the correct page DOM to flip the page. So I added extra code to get the correct page DOM for the page flip. Can you please take a look if the code acceptable or not ?
Thanks

src/js/page.js
```diff
@@ -89,6 +89,22 @@ export default {
...
+                    if ( $input ) {
+                        this._flipTo( $sibling );
+                        $sibling = null;
+                    } else {
+                        $sibling = $sibling.previousElementSibling;
+                    }
+                } else {
+                    $sibling = $sibling.previousElementSibling;
+                }
+            }
```
MartijnR, line 107: This fix relies on placing comment questions as siblings of the questions they link to, which is recommended but not obligatory. 

I think the determination of refererName is too fragile, and may easily cause an exception.

Consider directly looking for the linked question, e.g with something like this:
```js
const e = $e[0];// we're trying to get rid of jQuery, so let's not add any more. Feel free to rewrite some additional existing usage in this function (e.g. let closest = e.closest('[role="page"]') etc.
...
else {
  // if $e is a comment question, and it is not inside a group, there is no closest page
   const referer = e.querySelector('input[data-for]');
   const ancestor = e.closest('.or-repeat, form.or');
   if (referer && ancestor){
     const linkedQuestion = ancestor.querySelector(`[name="${refer.dataset.for}"]`);
     if (linkedQuestion){
        this._flipTo($(linkedQuestion));
     }
  }
}
```

I also needs a clear test (for the regular comment widget, not the OC dn widget). Otherwise, this fix cannot go in Enketo Core. See test/forms/comment.xml for an example form that you can perhaps use as-is, or copy and augment.

MartijnR (changes requested):

src/js/page.js
theywa: @MartijnR ~ code updated and tested on enketo-express-oc. Is that acceptable?
Also, I'm not sure about the test case. I test the default one on the test directory(http://localhost:8005/?xform=comment.xml) it still works fine also runs npm test. 
you said 
> See test/forms/comment.xml for an example form that you can perhaps use as-is, or copy and augment.

you mean to add a new test on the widget.comment.spec.js, right? if yes, then what kind of test should I add?
Thanks for the help

MartijnR: I think the test should probably go in page.spec.js. There are different ways of doing this, but of course make sure the test fails without your fix.

One way is to call flipToPageContaining manually, and then check if the current page has been updated (or that _flipTo was called with the correct parameter).

theywa: got it, test-case added and I update the page.js by removing the input tag on query selector because if the user sets the input comment appearance to multiline then it will be generated as textarea, not input so the selector will be failed if updated.

src/js/page.js
```diff
@@ -81,14 +81,22 @@ export default {
      * @param {jQuery} $e - Element on page to flip to
      */
     flipToPageContaining( $e ) {
+        const e = $e[ 0 ];
         let $closest;
```
MartijnR, line 85: No need to define here. It's also no longer variable so can be declared with `const = e.cl..` directtly.

src/js/page.js
```diff
@@ -81,14 +81,22 @@ export default {
...
-
-        $closest = $e.closest( '[role="page"]' );
-        $closest = ( $closest.length === 0 ) ? $e.find( '[role="page"]' ) : $closest;
-
-        // If $e is a comment question, and it is not inside a group, there may be no $closest.
-        if ( $closest.length ) {
-            this._flipTo( $closest[ 0 ] );
+        $closest = e.closest( '[role="page"]' );
```
MartijnR, line 86: We are using the convention to define jQuery objects with a $ prefix. This is not a jQuery object anymore, so it should be `closest` to avoid confusion.

MartijnR (changes requested):
Thanks! Just a minor thing. It looks fine otherwise.

test/spec/page.spec.js
```diff
@@ -31,4 +31,23 @@ describe( 'Pages mode', () => {
...
+            form.init();
+
+            const currentPage = form.pages.current;
+            const pageTwoTextAreaHiddenComment = form.view.html.querySelector( '[data-for="/data/item2"]' );
+            const pageComment = form.view.html.querySelector( 'input[name="/data/item2"]' ).closest( '[role="page"]' );
+            const pageTwoTextAreaCommentAncestor = pageTwoTextAreaHiddenComment.closest( '[role="comment"]' );
+    
+            form.pages.flipToPageContaining( [ pageTwoTextAreaCommentAncestor ] );
+            expect( currentPage ).not.toEqual( form.pages.current );
+            expect( pageComment ).toEqual( form.pages.current );
```
MartijnR, line 47: Travis testing fails because in the latest master we moved away from Jasmine to Chai (so `not.to.equal` and `to.equal`, I think)

MartijnR (changes requested):
one more thing

src/js/page.js
theywa: code updated

src/js/page.js
theywa: code updated

test/spec/page.spec.js
theywa: merged the latest master for testing the changes and code updated

PR #797, opened 2021-07-19
Switch build from Rollup to esbuild

eyelidlessness:
- Moves /build/index.html to /src to properly include in build step and ignore /build
- Like enketo/enketo-express#289, switches from Rollup to esbuild for consistency and improved build performance


lognaturel (approved):

src/index.html
```diff
@@ -86,7 +86,7 @@
     <!-- <script src="//localhost:35729/livereload.js"></script> -->
 
     <!-- all the scripts -->
-    <script id="main-script" defer module type="text/javascript" src="./js/enketo-bundle.js" charset="utf-8"></script>
+    <script id="main-script" defer module type="text/javascript" src="./js/app.js" charset="utf-8"></script>
```
lognaturel, line 89: @MartijnR, when you get a moment, would be great to get a sanity check on this change and whether you know of contexts in which it could cause problems.

This is a companion to https://github.com/enketo/enketo-express/pull/289 and I'm merging now after letting it sit for a few days to avoid conflicts and start getting the benefits to development.

Issue #795, opened 2021-07-13
test coverage reported as 100%


MartijnR:
I'm assuming that is incorrect, but haven't actually checked if we got that amazing improvement.

lognaturel:
Unfortunately not! We've decided to track coverage on a per-PR basis and remove the coverage badge so the build goes faster. #801

PR #798, opened 2021-07-23
Upgrade Node & NPM dependencies, fix NPM audit vulnerabilities

eyelidlessness:
### Changes

- Upgraded `enketo-transformer` temporarily to `https://github.com/eyelidlessness/enketo-transformer.git#node-16-dependency-upgrades`, this change will depend on a transformer release
- Pin engines to Node >=14 <17, NPM >=6 <7
- Pin Volta to Node 16/NPM 6
- Set up matrix testing for Travis CI to use NPM 14 and 16
- Added `preinstall` script to detect when local install includes `libxslt@0.9.0`, if so run `npm ci` to ensure all previously compiled binaries are upgraded for the current Node version
- Upgraded [sub-]dependencies for all outstanding security vulnerabilities
- `npm update` for dependencies with semver-compatible upgrades
- Upgraded other outdated dependencies:
  - `eslint-plugin-jsdoc`
  - `html5sortable`
- Removed broken code coverage badge (see #795, but even updating the badge was broken by the upgrade)

### Validation steps taken

- `npm clean-install`
- `npm install` with previous `node_modules` in place
- `npm audit` shows only vulnerabilities which cannot currently be patched, which are:
  - `publish-please > inquirer > lodash` & `publish-please > lodash`: dependencies are pinned to this specific version and cannot be updated to the semver-compatible fix
  - `mergexml > xmldom`: the dependency is `0.y.z` and the fix is treated as semver-incompatible
- `npm outdated` shows only false positives:
  - `enketo-transformer`
  - `leaflet-draw`
  - `signature_pad`
- `npm run test` passes (this also includes `npm run eslint-check`)
- `npm run start` + manually view a form
- `npx grunt` before/after, with the same logic used in `enketo-core` to transform all the test fixtures, to verify they don't produce a diff, further validating changes in transformer
- `npm run build-docs` before/after doesn't produce a diff, other than those expected from substantive changes:
  - `docs/Form.html` and `docs/js_form_js.html` reflect the transformer version change
  - `docs/index.html` reflects removal of code coverage badge
- `npx grunt build` does not produce a diff, because there is no other substantive Node behavior in the project


eyelidlessness:
Did another pass in 5c0250f, eliminating the remaining vulnerabilities. It depends on two open PRs:

- `publish-please`: https://github.com/inikulin/publish-please/pull/100
- `js-merge-xml`: https://github.com/hareko/js-merge-xml/pull/19

In the event either of those are not merged/released, I'd prefer we either continue to depend on my fork or roll back, rather than delay these upgrades further.

I've performed all of the same validations.

yanokwa (approved):
I ran `npm audit`, `npm outdated` and checked Dependabot for vulns. I also confirmed `npm run test` passed and I could view a form with `npm run start` and all widgets form looked fine. Ditto with 

GitHub is reporting a [vulnerability](https://github.com/advisories/GHSA-6fc8-4gx4-v693) that we won't be fixing.  Per @eyelidlessness's rationale, 

> i don't see any way to resolve `ws`. this a subdep of `grunt-contrib-connect`. i think we're going to find more and more grunt-related things are basically abandonware. that said, i don't think we're actually at risk from this particular vulnerability, so i'm not in a rush to try to find an alternative solution. the grunt plugin is only in use for running local dev server, so we'd have to exploit ourselves to actually be at risk

Issue #373, opened 2016-06-08
add warning to changelog 4.7.0


MartijnR:
to use [enketo-transformer 1.8.0+](https://github.com/enketo/enketo-transformer/blob/master/CHANGELOG.md#180---2016-05-04)

MartijnR:
#372

PR #803, opened 2021-08-20
To fix if the user make an annotate and click on hide-canvas-btn with…

jdugh:
…in DELAY

To avoid any confilct between the updatevalue trigger by forceupdate and the futur updatevalue on canvas reload and potential another async event

### DIFF
--- a/src/widget/draw/draw-widget.js
+++ b/src/widget/draw/draw-widget.js
@@ -182,7 +182,9 @@ class DrawWidget extends Widget {
                         that.$widget.removeClass( 'full-screen' );
                         that.pad.off();
                         that._forceUpdate();
-                        that._resizeCanvas( canvas );
+                        // To fix if the user make an annotate and click on hide-canvas-btn within DELAY :
+						// To avoid any confilct between the updatevalue trigger by forceupdate and the futur updatevalue on canvas reload and potential another async event
+						setTimeout( that._resizeCanvas.bind(that, canvas ), DELAY);
 
                         return false;
                     } ).click();


MartijnR:
Thanks for the PR. It is not clear to me exactly what issue this is addressing (though I read the commit message and comments, and tried a few things). 

1. Is the issue occurring on touchscreen devices only?
2. How can we reproduce it? 
3. Can we reproduce it with the annotate widget in this form: https://enke.to/widgets?

jdugh:
Sorry, I did the PR a bit quickly without explanation.
On mobile phone (Samsung A51 - Chrome - With inspect devices for the console).
I have found a problem with the annotate question : 
On https://enke.to/widgets?, I go to the annotate question.
- I click on "Draw/Sign" (full-screen mode)
- I upload a photo
- I draw one line
- I click on previous (hide-canvas-btn) before 1,5 seconde
- I submit my form

If I add the photo and click on previous (hide-canvas-btn) before 1,5 seconde, same problem

The photo don't have the right name between [f.name and filename](https://github.com/enketo/enketo-express/blob/98f7412d877229fbae8d310f056919aecb2c27eb/public/js/src/module/connection.js#L225) - 1 seconde deffirence -  and I get the console.error (Error occured when trying to retrieve annotation-12_2_3.png) and the popup error (Does this popup appear to be brand new by the way? - great !)

In the draw widget I feel there is some async methode, I find it hard to understand where the real problem is coming from.

If you are an another way/idea for solve the problem, I can change my PR (and make the test).
Maybe we can decrease the DELAY, but I don't think if it's a sustainable solution.

Thank's for all !

Screenshot_20210821-120219_Chrome
Screenshot_20210821-120236_Chrome
Capture

MartijnR:
Thank you for the explanation. I was able to reproduce with Chrome on Android upon submission. 

This may have something to do with storing the updated drawing in the browser database or a deliberate delay in the drawing widget. It would be good to figure out (in Enketo Express or Enketo Core) what code is responsible for this.

Adding a delay like this should be more of a last resort workaround in my opinion (though we have a few of those solutions already, I think...).

lognaturel:
@eyelidlessness This is at least tangentially related to https://github.com/enketo/enketo-express/issues/339

eyelidlessness:
@lognaturel yes this does look related! And I'm looking at addressing the feedback in enketo/enketo-express#373 right now, so I'll try to make sure I account for this as well.

lognaturel:
@jdugh can you please see whether https://github.com/enketo/enketo-express/pull/373 addresses the issue you had identified?

jdugh:
Thank you all. I close this PR. It's corrected by [#373](https://github.com/enketo/enketo-express/pull/373) - Release 3.0.5.

PR #806, opened 2021-08-27
[WIP] Fix nested repeats with relevance

eyelidlessness:
This is, temporarily, a targeted fix for one user, addressing a case where relevance on a nested/grouped repeat only works for the first top-level repeat. It causes a test failure for another use case, which provides helpful testing context and sort of hints at a likely direction for a more general fix.


lognaturel:
Replaced by #807

PR #807, opened 2021-08-30
[Fix] Ensure nested repeats are shown when relevance is true

### LINKED ISSUE #804: No relevance evaluated on 2nd iteration of a nested repeat
**Describe the bug**
Enketo v2.8.0-11-g98f7412 does not evaluate relevance on the second iteration of a nested repeat.

**To Reproduce**
Given a form like this:
Screen Shot 2021-08-21 at 16 10 24

I expect each iteration of `field_asset_rpt` to show me the nested `documentation_rpt` and `instrumentation_rpt` repeats if I select yes for `documentation_exists` or `instrumentation_exists`. In the first iteration of field asset, it works great. In the second iteration, I don't see either documentation or instrumentation repeats.

The screenshot below shows the current behavior for just the instrumentation repeat. Green arrows are the first iteration which works great. Red arrows are the second iteration which don't work.

Screen Shot 2021-08-21 at 16 12 31

I've included test XLSForm and XML at [enketo-repeat-bug.zip](https://github.com/enketo/enketo-core/files/7026331/enketo-repeat-bug.zip). The XLSForm was converted at https://getodk.org/xlsform and XML was prettified to make the XML easier for humans to read.

**Expected behavior**

 - OS: macOS 11.5.1
 - Browser: Chrome Version 92.0.4515.159
 
**Additional context**
Works great in Collect.

eyelidlessness:
Fixes #804. This is branched off of master, unlike #806 which we should likely close.

Some of these tests are likely redundant, but it was hard for me to be sure, so I felt more comfortable adding them.

### DIFF
--- a/src/js/relevant.js
+++ b/src/js/relevant.js
@@ -87,8 +87,15 @@ export default {
              * but currently has 0 repeats, the context will not be available. This same logic is applied in output.js.
              */
             let context = p.path;
-            if ( getChild( node, `.or-repeat-info[data-name="${p.path}"]` ) && !getChild( node,  `.or-repeat[name="${p.path}"]` ) ) {
-                context = null;
+
+            const repeatInfo = getChild( node, `.or-repeat-info[data-name="${p.path}"]` );
+
+            if ( repeatInfo != null && !getChild( node, `.or-repeat[name="${p.path}"]` ) ) {
+                const count = this.form.repeats.updateViewInstancesFromModel( repeatInfo );
+
+                if ( count === 0 ) {
+                    context = null;
+                }
             }
 
             /*
--- a/src/index.html
+++ b/src/index.html
@@ -86,7 +86,7 @@
     <!-- <script src="//localhost:35729/livereload.js"></script> -->
 
     <!-- all the scripts -->
-    <script id="main-script" defer module type="text/javascript" src="./js/app.js" charset="utf-8"></script>
+    <script id="main-script" defer module type="text/javascript" src="./app.js" charset="utf-8"></script>
 
 </body>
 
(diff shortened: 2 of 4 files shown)


MartijnR:
Thanks for tackling a nested repeat issue! I am committing to doing a review.

src/js/relevant.js
```diff
@@ -87,8 +87,15 @@ export default {
              * but currently has 0 repeats, the context will not be available. This same logic is applied in output.js.
              */
             let context = p.path;
-            if ( getChild( node, `.or-repeat-info[data-name="${p.path}"]` ) && !getChild( node,  `.or-repeat[name="${p.path}"]` ) ) {
-                context = null;
+
+            const repeatInfo = getChild( node, `.or-repeat-info[data-name="${p.path}"]` );
+
+            if ( repeatInfo != null && !getChild( node, `.or-repeat[name="${p.path}"]` ) ) {
+                const count = this.form.repeats.updateViewInstancesFromModel( repeatInfo );
```
MartijnR, line 94: I believe the underlying issue is that the nested repeats are not created when the top-level repeat (no. 2) is created (after the user clicks the plus button). 

The relevant module should not be responsible for fixing that, as it's the responsibility of the repeat module. It's odd that something so elementary is not done correctly. I wonder if this bug was introduced when we started allowing 0 repeat instances. There seems to be some (still mysterious) special condition in the test form, because in e.g. http://localhost:8005/?xform=nested_repeats.xml the bug doesn't surface.

MartijnR (changes requested):

src/js/relevant.js
eyelidlessness: I was hesitant to apply this fix here too, I'm glad you agree. To be honest I chose this path because I struggled to understand the repeat initialization flow and wasn't sure how to address it there. But I'll take another look because I agree it would be better solved there.

PR #809, opened 2021-09-03
Prepare for 6.0.0

yanokwa:
Todos
- [x] `npx publish-please --dry-run` does not work
- [ ] `npm run test-browsers` fails. See https://github.com/enketo/openrosa-xpath-evaluator/issues/133.


eyelidlessness:
I think I know why `npx publish-please --dry-run` isn't working, and it's my fault. 🙃 

With the dependency pinned to my fork, `publish-please` itself has not yet been built and published with... `publish-please`. So I'm going to publish my fork in my own scope for now as `@eyelidlessness/publish-please`, and validate that that works which I strongly suspect it will.

yanokwa:
@MartijnR Do you usually run `npm run test-browsers` before a release of core?

MartijnR:
No, I think I have been forgetting to do that for the past few years.... 😳 I'm guessing it's not looking good?

tutorials/20-development.md
```diff
@@ -16,7 +16,7 @@
 * When creating new functions/Classes, make sure to describe them with JSDoc comments.
 * JavaScript style see [ESLint](./eslintrc.json) config files. The check is added to the grunt `test` task. You can also manually run `grunt eslint:fix` to fix style issues.
 * Testing is done with Mocha and Karma (all: `grunt karma`, headless: `grunt karma:headless`, browsers: `grunt karma:browsers`)
-* Tests can be run in watch mode for [TDD](https://en.wikipedia.org/wiki/Test-driven_development) workflows with `npm run test-watch`, and support for debugging in [VSCode](https://code.visualstudio.com/) is provided. For instructions see [./#debugging-test-watch-mode-in-vscode](Debugging test watch mode in VSCode) below
+* Tests can be run in watch mode for [TDD](https://en.wikipedia.org/wiki/Test-driven_development) workflows with `npm run test-watch`, and support for debugging in [VSCode](https://code.visualstudio.com/) is provided. For instructions see [./#debugging-test-watch-mode-in-vscode](debugging-test-watch-mode-in-vscode) below
```
eyelidlessness, line 19: I think this should actually be 

```suggestion
* Tests can be run in watch mode for [TDD](https://en.wikipedia.org/wiki/Test-driven_development) workflows with `npm run test-watch`, and support for debugging in [VSCode](https://code.visualstudio.com/) is provided. For instructions see [Debugging test watch mode in VSCode](./#debugging-test-watch-mode-in-vscode) below
```

Markdown is fun 🙃

tutorials/20-development.md
```diff
@@ -53,11 +53,14 @@ For custom themes that go beyond just changing colors and fonts, keep in mind al
...
+1. Check [Dependabot alerts](https://github.com/enketo/enketo-transformer/security/dependabot) for vulnerabilities
+2. Update dependencies: `npm update`. If enketo-transformer is updated, bump the version in `src/js/form.js`
+3. `npm audit fix`
+4. Make sure tests pass: `npm run test`.
+5. Beautiful code: `npm run beautify`.
+6. Build documentation: `npm run build-docs`.
+7. Bump the version tag in `package.json` file (we follow [semantic versioning](https://semver.org/)). Bump to major version if downstream has to make changes.
+8. Merge all your changes to `master` (through PR).
+9. Add git tag of new version.
+10. Publish module to NPM: `npm run`
```
eyelidlessness, line 65: Probably should be:

```suggestion
10. Publish module to NPM: `npm publish`
```

eyelidlessness (changes requested):

eyelidlessness:
> No, I think I have been forgetting to do that for the past few years.... 😳 I'm guessing it's not looking good?

I’ll look into it in a few. I suspect something may have broken in some of the tooling changes recently.

We should also see if we can get it running in CI so we catch it sooner next time 😬

eyelidlessness (approved):
I think all of the browser test failures are related to the issues discussed in enketo/openrosa-xpath-evaluator#133. Since these all predate the changes we intend to release here, I think we can go ahead with this and address those test failures and any other bugs we identify after release.

package-lock.json
```diff
@@ -1,39 +1,8757 @@
 {
     "name": "enketo-core",
-    "version": "5.18.1",
-    "lockfileVersion": 1,
+    "version": "6.0.0",
+    "lockfileVersion": 2,
```
yanokwa, line 4: @eyelidlessness Can you sanity check that this is fine?

eyelidlessness: I don't think this is right. It suggests the lockfile was generated with NPM 7, which will cause installs to fail. We saw this previously with libxmljs and libxslt, but I see it is also failing with the git reference to my mergexml fork as well.

Are you using Volta? If not, I'd highly recommend it as it will (quickly, unlike nvm) use the Node/NPM versions pinned in package.json.

eyelidlessness (approved):

PR #810, opened 2021-09-08
[fix] Decimal fields accept any dot (.) or comma (,) combinations as inputs but submit empty

### LINKED ISSUE #792: Decimal fields accept any combination of , . - as inputs but submit empty
**Describe the bug**
Field type `decimal` accepts values with thousands seperators (like 123.123.123) but will not submit any value to the ODK server (field arrives empty and is shown as empty in e.g. ODK Central). The field also accepts wild . and , combinations like `12.212,12` or `123,123.12` or even `123,.,123`. Enketo does not consider this as invalid or warn anyhow about it and all these combinations will arrive empty at the server.

This seems to be a Firefox peculiarity. Still, I think this should be considered as invalid by Enketo.
See this warning on MDM: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#using_number_inputs

See some combinations in attached screenshot (my locale uses comma as decimal seperator).
grafik

**To Reproduce**
See above.

**Expected behavior**
Only decimals with a single "." or "," as decimal seperator should be considered as valid. For other values, the field should visibally be invalid.

 - Enketo 2.7.3
 - Latest Firefox 89 on Windows 10
 - Chrome does not seem to be affected

**Additional context**
Find minimal XlsForm for tests attached.
[decimal-seperator-test.xlsx](https://github.com/enketo/enketo-core/files/6663846/decimal-seperator-test.xlsx)

mattelacchiato:
Fixes #792 
It seems that the blur events need to be added to each input element. Therefore, I've chosen to use the `$.on` function for adding the event listeners.

Didn't added any tests, since we won't be able to add tests: The internal state is already empty for invalid values, so we have no chance to grab the displayed value in a testing scenario.

### DIFF
--- a/src/js/mask.js
+++ b/src/js/mask.js
@@ -57,23 +57,22 @@ export default {
             }
 
         } );
+
         /*
-     * Workaround for most browsers keeping invalid numbers visible in the input without a means to access the invalid value.
-     * E.g. see https://bugs.chromium.org/p/chromium/issues/detail?id=178437&can=2&q=178437&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified
-     *
-     * A much more intelligent way to solve the problem would be to add a feedback loop from the Model to the input that would
-     * correct (a converted number) or empty (an invalid number). https://github.com/enketo/enketo-core/issues/407
-     */
-        form.addEventListener( 'blur', event => {
-            if ( event.target.matches( selector ) ){
+         * Workaround for most browsers keeping invalid numbers visible in the input without a means to access the invalid value.
+         * E.g. see https://bugs.chromium.org/p/chromium/issues/detail?id=178437&can=2&q=178437&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified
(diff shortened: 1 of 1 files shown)


src/js/mask.js
```diff
@@ -57,23 +57,22 @@ export default {
...
-     */
-        form.addEventListener( 'blur', event => {
-            if ( event.target.matches( selector ) ){
+         * Workaround for most browsers keeping invalid numbers visible in the input without a means to access the invalid value.
+         * E.g. see https://bugs.chromium.org/p/chromium/issues/detail?id=178437&can=2&q=178437&colspec=ID%20Pri%20M%20Stars%20ReleaseBlock%20Component%20Status%20Owner%20Summary%20OS%20Modified
+         *
+         * A much more intelligent way to solve the problem would be to add a feedback loop from the Model to the input that would
+         * correct (a converted number) or empty (an invalid number). https://github.com/enketo/enketo-core/issues/407
+         */
+        this.form.view.$.on( 'blur', `input${selector}:not(.ignore)`, event => {
```
vituslehner, line 68: Hi @mattelacchiato hope you're doing well! Can you give me a little hint about what has changed in your PR? I can mainly see a change in how the event listener gets added in line 68, but not that any other logic has changed.

PR #811, opened 2021-09-08
fixed: isValid method always returns true

MartijnR:
Discovered a bug in the code while writing some tests for OC extensions to enketo-core.

* [x] Martijn still to describe the actual real-life issue caused by this bug (something with page validation).

### DIFF
--- a/src/js/form.js
+++ b/src/js/form.js
@@ -882,7 +882,7 @@ Form.prototype.setInvalid = function( control, type = 'constraint' ) {
     // and clicks the Next or Previous button without having left that form control (no blur event)
     // we temporarily block the navigation action to ensure the user sees the validation error message.
     // The user will have to click again to navigate.
-    if ( config.validatePage === false && this.isValid( control ) ) {
+    if ( config.validatePage === false ) {
         this.blockPageNavigation();
     }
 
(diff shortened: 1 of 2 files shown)


MartijnR:
No issue, just obsolete code because `isValid` is always true due to sloppy coding. See inline comment.

src/js/form.js
```diff
@@ -923,24 +927,6 @@ Form.prototype.blockPageNavigation = function() {
...
- * @param {Element} node - form control HTML element
- * @return {!boolean} Whether the question/form is not marked as invalid.
- */
-Form.prototype.isValid = function( node ) {
-    const invalidSelectors = [ '.invalid-required', '.invalid-relevant' ].concat(  this.constraintClassesInvalid.map( cls => `.${cls}` ) );
-    if ( node ) {
-        const question = this.input.getWrapNode( node );
-        const cls = question.classList;
-
-        return !invalidSelectors.some( selector => cls.contains( selector ) );
```
MartijnR, line 938: This always returns true because `selector` should a className without a dot prefixed to it.

src/js/form.js
```diff
@@ -923,24 +927,6 @@ Form.prototype.blockPageNavigation = function() {
...
-    }
-
-    return !this.view.html.querySelector( invalidSelectors.join( ', ' ) );
```
MartijnR, line 941: I suspect this code is never reached but I am not 100% sure.

Issue #816, opened 2021-09-24
Video and audio keep playing when navigating to other pages in the form


latin-panda:
**Describe the bug**
At Medic we got a [bug report ](https://github.com/medic/cht-core/issues/6589) where users can still hear the video/audio playing when navigating to other pages in the form. 
This is a very common issue for `<video>` and `<audio>` elements when the container has `display: none` as the elements still exists in the DOM.

**To Reproduce**
1. Add embed .mp4 and mp3 in a form that has multiple pages. Sample:
```xml
    <model>
      <itext>
        <translation lang="en">
          <!-- video and audio are form attachments -->
          <text id="somevideo">
            <value form="video">jr://video.mp4</value>
          </text>
          <text id="someimage">
            <value form="image">jr://image.jpg</value>
          </text>
          <text id="someaudio">
            <value form="audio">jr://audio.mp3</value>
          </text>
        </translation>
      </itext>
      <instance>
        <media id="multimedia">
          <meta>
            <instanceID/>
          </meta>
        </media>
      </instance>
    </model>

.........
.........

<group appearance="field-list" ref="g">
  <input ref="q1">
    <label ref="jr:itext('somevideo')"/>
  </input>
</group>

<group appearance="field-list" ref="g">
  <input ref="q2">
    <label ref="jr:itext('someaudio')"/>
  </input>
</group>

<group appearance="field-list" ref="g">
  <input ref="q3">
    <label ref="jr:itext('someimage')"/>
  </input>
</group>
```
2. Play the video/audio and navigate to next/prev page.
3. The video/audio continues to play.

**Expected behavior**
The video/audio should stop playing when viewing other pages. 

**Browser and OS (please complete the following information):**
 - Browser: Chrome

latin-panda:
I've sent [this PR](https://github.com/enketo/enketo-core/pull/817) as suggestion to fix this issue.

If this is okay, please consider backport it to 5.18.X, Medic still support projects in Node 8 to 12 and won't be able to meet Enketo v6.0.0 min engine for a while.
Thanks

PR #817, opened 2021-09-24
Pause multimedia when switching pages

latin-panda:
In this PR:
- Pause multimedia (video and audio) from current page before navigating to next/prev page.

This code is a suggestion to fix [this issue](https://github.com/enketo/enketo-core/issues/816).

If this is okay, please consider backport it to 5.18.X, we noticed Enketo version 6.0.0 has a minimum of Node 14, at Medic we still support projects in Node 8 to 12 and won't be able to upgrade Enketo to 6.0.0 for a while.


latin-panda:
I @eyelidlessness wondering if this is okay, thanks!

eyelidlessness (approved):
I have manually validated that this works as expected. I also threw together a quick test, which I'll follow up with later.

Thanks!

PR #820, opened 2021-10-01
Fix name collision error in repeat

### LINKED ISSUE #815: Name collision with calculate in repeat causes error
**Describe the bug**
This is a bit of an edge case, but we are seeing an issue where an error is thrown if you have a `calculate` field in a sub-group within a `repeat` and the `calculate` has the _same name_ as a prior question in the `repeat` (but the question is not in the same sub-group as the `calculate`).  When a form is setup like this, the error gets thrown if there is _more than one_ entry in the repeat.

Error Log:

```
form-model.js:1239 no context element found /data/outside_group/test_repeat/inner_group/text_in_repeat -1
FormModel.evaluate @ form-model.js:1239
form-model.js:1318 Uncaught FormLogicErrormessage: "Could not evaluate: /model/instance[1]/data/outside_group/repeat_count, message: Failed to execute 'evaluate' on 'XPathEvaluator': parameter 2 is not of type 'Node'."name: "FormLogicError"stack: "Error\n    at new FormLogicError (https://enketo.getodk.org/js/build/enketo-webform.js:47547:18)\n    at FormModel.evaluate (https://enketo.getodk.org/js/build/enketo-webform.js:48699:15)\n    at Object._updateCalc (https://enketo.getodk.org/js/build/enketo-webform.js:56597:59)\n    at https://enketo.getodk.org/js/build/enketo-webform.js:56505:18\n    at Array.forEach (<anonymous>)\n    at Object.update (https://enketo.getodk.org/js/build/enketo-webform.js:56488:13)\n    at https://enketo.getodk.org/js/build/enketo-webform.js:57175:12\n    at Array.forEach (<anonymous>)\n    at HTMLFormElement.<anonymous> (https://enketo.getodk.org/js/build/enketo-webform.js:57174:30)\n    at Object.add (https://enketo.getodk.org/js/build/enketo-webform.js:49372:15)"[[Prototype]]: Error
```

**To Reproduce**

1. Open [this xForm](https://docs.google.com/spreadsheets/d/1cCXefohzrcGmiKTXBaOS5xPOl3W9ORN5/edit?usp=sharing&ouid=110810027707570331679&rtpof=true&sd=true)
1. Enter `2` (or any value > 1) for the "How many times to repeat" question
1. See the error in the browser console

One definite side-effect that we have noticed is that this error affects the `note` widgets within the `repeat` and causes the notes in subsequent repeat entries (all but the first) to also display a read-only text box.  You can see in the screenshot that in the second entry of the repeat, the note, "This is a repeated note" is actually displayed as a read-only text question.

image

**Expected behavior**

Since the two fields with the same name are not in the same group (and so actually have distinct paths) I expected this to work without an error.  

One practical use case that could result in this name collision is found in the [default `delivery` form](https://github.com/medic/cht-core/blob/master/config/default/forms/app/delivery.xlsx) in cht-core. This form has several questions within a repeat and then uses calculates within a sub-group in the repeat to massage the data from the questions. However, the calculates in the sub-group just have the same names as the questions.

This seems to be an issue in how the index is getting caluculated to pick which data node to update: https://github.com/enketo/enketo-core/blob/4347aa79485db1bc281d867ecc7a6d3294ede149/src/js/calculate.js#L61
That line of code is simply grabbing the first child element of the repeat entry that has the `dataNodeName` and then using that element to determine the actual index to use: https://github.com/enketo/enketo-core/blob/4347aa79485db1bc281d867ecc7a6d3294ede149/src/js/calculate.js#L62
The problem is that with the name collision, the element selected as the `dataNode` is not the calculate (but is instead the text question with the same name).  This causes `props.index` to be set to `-1` which, in turn, eventually triggers the error seen in the logs.

Updating the code for selecting the `dataNode` to the following solved the issue for me:

```.js
const childNodeList = this.form.model.node( updated.repeatPath, updated.repeatIndex ).getElement().querySelectorAll( dataNodeName );
const dataNode = Array.from(childNodeList).filter( node => dataNodes.includes( node ) )[0];
```

Let me know if this explanation/analysis makes sense and if the proposed solution is acceptable.  I would be happy to submit this in a PR.  Thank you!

 - OS: Linux Mint 20.2
 - Browser: Brave 1.29.79 Chromium: 93.0.4577.63

jkuester:
## Details

Updates the the processing logic for `calculate` fields inside of repeats so that the index gets properly selected even if there are other questions in the same repeat (but not in the same sub-group) with the same name. More details about the issue and how to recreate it are documented in #815.

The change in logic comes when selecting the child nodes of the repeat entry by the `dataNodeName`. Previously, just the first node that was found with a matching name was used. However, this might not be the actual `calculate` node. So, I have updated the logic to filter all the child nodes found with the `dataNodeName` down to just the one that is also present in the `dataNodes` array (with all the instances of that `calculate` field for the repeat). Once the proper `dataNode` is resolved, the index is calculated correctly and the rest of the processing can proceed without error.

## Tests

I am happy to add a new form/test that is specific to this issue if you would like. However, I was able to get test coverage of this issue from the existing tests simply by updating the name of the `calculate` in the `calcs_in_repeats.xml` form to match the question in the repeat.  When I ran the `form.spec.js` tests with the new calculate name, but without the implementation code changes, all of the tests that use the `calcs_in_repeats.xml` form failed as expected with the following error:

```
	Error: Uncaught FormLogicError: Could not evaluate: /model/instance[1]/calcs_in_repeats/rep1/num1  * 20, message: Failed to execute 'evaluate' on 'XPathEvaluator': parameter 2 is not of type 'Node'. (/tmp/08991e993b53562045b647f27beae811-bundle.js:37302)
	    at Object.add (/home/jlkuester7/git/enketo-core/src/js/repeat.js:326:19 <- /tmp/08991e993b53562045b647f27beae811-bundle.js:38070:15)
	    at HTMLButtonElement.<anonymous> (/home/jlkuester7/git/enketo-core/src/js/repeat.js:99:18 <- /tmp/08991e993b53562045b647f27beae811-bundle.js:37951:14)
	    at HTMLFormElement.dispatch (/home/jlkuester7/git/enketo-core/node_modules/jquery/dist/jquery.js:5430:27 <- /tmp/08991e993b53562045b647f27beae811-bundle.js:3344:105)
	    at HTMLFormElement.elemData.handle (/home/jlkuester7/git/enketo-core/node_modules/jquery/dist/jquery.js:5234:28 <- /tmp/08991e993b53562045b647f27beae811-bundle.js:3232:118)
	    at Object.trigger (/home/jlkuester7/git/enketo-core/node_modules/jquery/dist/jquery.js:8719:12 <- /tmp/08991e993b53562045b647f27beae811-bundle.js:5288:24)
	    at HTMLButtonElement.<anonymous> (/home/jlkuester7/git/enketo-core/node_modules/jquery/dist/jquery.js:8797:17 <- /tmp/08991e993b53562045b647f27beae811-bundle.js:5334:29)
	    at Function.each (/home/jlkuester7/git/enketo-core/node_modules/jquery/dist/jquery.js:385:19 <- /tmp/08991e993b53562045b647f27beae811-bundle.js:770:30)
	    at jQuery2.fn.init.each (/home/jlkuester7/git/enketo-core/node_modules/jquery/dist/jquery.js:207:17 <- /tmp/08991e993b53562045b647f27beae811-bundle.js:658:28)
	    at jQuery2.fn.init.trigger (/home/jlkuester7/git/enketo-core/node_modules/jquery/dist/jquery.js:8796:15 <- /tmp/08991e993b53562045b647f27beae811-bundle.js:5333:25)
```
Once the implementation code changes were applied, the test errors were resolved.

Closes #815

### DIFF
--- a/src/js/calculate.js
+++ b/src/js/calculate.js
@@ -58,7 +58,8 @@ export default {
                      * and the dependent node is inside the same repeat, we can prevent the expensive index determination
                      */
                     const dataNodeName = ( name.lastIndexOf( '/' ) !== -1 ) ? name.substring( name.lastIndexOf( '/' ) + 1 ) : name;
-                    const dataNode = this.form.model.node( updated.repeatPath, updated.repeatIndex ).getElement().querySelector( dataNodeName );
+                    const childNodeList = this.form.model.node( updated.repeatPath, updated.repeatIndex ).getElement().querySelectorAll( dataNodeName );
+                    const dataNode = Array.from(childNodeList).filter( node => dataNodes.includes( node ) )[0];
                     props.index = dataNodes.indexOf( dataNode );
                     this._updateCalc( control, props, emptyNonRelevant );
                 } else if ( control.type === 'hidden' ) {
--- a/test/forms/calcs_in_repeats.xml
+++ b/test/forms/calcs_in_repeats.xml
@@ -15,7 +15,8 @@
             <num1/>
             <grp>
               <note/>
-              <calc3/>
+              <!-- calculate has the same name ("num1") as the int question above to test for name collisions in repeats: https://github.com/enketo/enketo-core/issues/815 -->
+              <num1/>
             </grp>
           </rep1>
           <note1/>
@@ -29,7 +30,7 @@
       <bind nodeset="/calcs_in_repeats/cond1" type="string"/>
       <bind nodeset="/calcs_in_repeats/rep1/num1" type="int"/>
       <bind nodeset="/calcs_in_repeats/rep1/grp" relevant=" /calcs_in_repeats/rep1/num1 > 5 "/>
-      <bind calculate=" /calcs_in_repeats/rep1/num1  * 20" nodeset="/calcs_in_repeats/rep1/grp/calc3" type="string"/>
+      <bind calculate=" /calcs_in_repeats/rep1/num1  * 20" nodeset="/calcs_in_repeats/rep1/grp/num1" type="string"/>
       <bind nodeset="/calcs_in_repeats/note1" readonly="true()" type="string"/>
       <bind calculate="concat('uuid:', uuid())" nodeset="/calcs_in_repeats/meta/instanceID" readonly="true()" type="string"/>
       <bind calculate="'47eed988240249948e42aadfa6bcdb86'" nodeset="/calcs_in_repeats/formhub/uuid" type="string"/>
--- a/test/spec/form.spec.js
+++ b/test/spec/form.spec.js
@@ -304,8 +304,8 @@ describe( 'calculations', () => {
         form.view.$.find( '.add-repeat-btn' ).click();
         form.view.$.find( '[name="/calcs_in_repeats/rep1/num1"]:eq(0)' ).val( '10' ).trigger( 'change' );
         form.view.$.find( '[name="/calcs_in_repeats/rep1/num1"]:eq(1)' ).val( '20' ).trigger( 'change' );
-        expect( form.model.node( '/calcs_in_repeats/rep1/grp/calc3', 0 ).getVal() ).to.equal( '200' );
-        expect( form.model.node( '/calcs_in_repeats/rep1/grp/calc3', 1 ).getVal() ).to.equal( '400' );
+        expect( form.model.node( '/calcs_in_repeats/rep1/grp/num1', 0 ).getVal() ).to.equal( '200' );
+        expect( form.model.node( '/calcs_in_repeats/rep1/grp/num1', 1 ).getVal() ).to.equal( '400' );
     } );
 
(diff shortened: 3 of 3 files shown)


eyelidlessness:
Thanks for submitting a PR! I'm taking a look at this now.

eyelidlessness (approved):
I can validate this does fix the issue, thanks @jkuester! I wasn’t especially familiar with this area of functionality, so I took some time to walk through it thoroughly. In the course of that I discovered another related issue, which may have a similar root cause.

I think I may be able to find a more general solution, but this is a good starting point for that, and I don’t want to hold up this fix for the general case.

I’ll write up a follow up issue for the other bug I found in a bit. But this looks good to go, thanks again!

PR #822, opened 2021-10-01
Fix/randomize 821

MartijnR:
Issue can be reproduced with http://localhost:8005/?xform=randomize.xml in master.

It is quite serious and it is surprising only one person has reported it so far. 

The fix is to used unordered snapshot results when evaluating user-defined XPath expressions. Out of an abundance of caution I kept some internal node-set XPath queries as ordered (though none of the tests failed).

It is possible that there is some use case (querying a secondary instance e.g.) that causes a problem, but I think it's quite unlikely, and considering the seriousness of this bug, I propose to go ahead with this fix unless a real issue can be identified of course.

### DIFF
--- a/src/js/form-model.js
+++ b/src/js/form-model.js
@@ -431,15 +431,15 @@ FormModel.prototype.mergeXml = function( recordStr ) {
             const instanceNode = that.node( path, 0 ).getElement();
             if ( instanceNode ) {
                 // TODO: after dropping support for IE11, we can also use instanceNode.children.length
-                if ( that.evaluate( './*', 'nodes', path, 0, true ).length === 0 ) {
+                if ( that.evaluate( './*', 'nodes-ordered', path, 0, true ).length === 0 ) {
                     // Select all text nodes (excluding repeat COMMENT nodes!)
-                    that.evaluate( './text()', 'nodes', path, 0, true ).forEach( node => {
+                    that.evaluate( './text()', 'nodes-ordered', path, 0, true ).forEach( node => {
                         node.textContent = '';
                     } );
                 } else {
                     // If the node in the default instance is a group (empty in record, so appears to be a leaf node
                     // but isn't), empty all true leaf node descendants.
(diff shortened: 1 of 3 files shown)


eyelidlessness:
I’m looking at this now, thanks @MartijnR!

eyelidlessness:
So it looks like there are two remaining uses of `'nodes'` rather than `'nodes-ordererd'` (which aren't visible in the PR so I can't comment inline):

1. `getRepeatCommentEl`, which then accesses the comment by subscript access with `repeatSeriesIndex`. It's unclear to me if this is intended to be used with `randomize`/fixed by this change. If it's not, I'm concerned about the potential for unstable order affecting repeats.
2. Within itemset.js `update`, which appears from the randomize.xml fixture to be the use case being addressed by this PR. But what about cases where `randomize` isn't in use? I'd have the same concern here.

I don't have enough context know how urgent this is/how to determine whether that risk is worth a quick fix. So I'm hesitant to approve or request changes at this time, I'll need some clarification on that. But I suspect the non-urgent-term solution should probably be in openrosa-xpath-evaluator, where `randomize()` would (presumably, along with any other order-altering functions) be implemented _after_ the initial XPath evaluation.

MartijnR:
thanks!

>1 
Good find. I'll update the use in getRepeatCommentEl. 

>2
Indeed, that the the use case we're addressing.

I also do not know for sure if unordered lists present a risk, but they likely are, in practice, ordered the same way as ordered lists. My limited testing confirms that the results are the same (only tested Chrome). The way I see it, there are 2 real use cases: querying secondary instances which will (always?) have fairly flat consistent repetitive data structures, and repeat questions in the primary instance which have a repetitive structure as well.

Yes, the alternative hacky solution Alex mentioned would be best done in the XPath evaluator to avoid duplicating (or exporting) the randomization code. The reason I (we probably) don't like it is that the evaluator does things correctly at the moment.

PR #831, opened 2021-10-20
Document release steps, add GH workflow

yanokwa:
One thing we should check is what browsers GH has installed.

Note the removal of Travis

test-browsers are failing due to these warnings.
```
Chrome 94.0.4606.81 (Linux x86_64) WARN: 'Data node: /thedata/nodeA with null-based index: 1 not found. Ignored.'
Chrome 94.0.4606.81 (Linux x86_64) ERROR: 'nodeset.setVal expected nodeset with one node, but received multiple'
```

The code in question is from src/js/nodeset.js:138, 143

```
    if ( targets.length > 1 ) {
        console.error( 'nodeset.setVal expected nodeset with one node, but received multiple' );

        return null;
    }
    if ( targets.length === 0 ) {
        console.warn( `Data node: ${this.selector} with null-based index: ${this.index} not found. Ignored.` );

```


eyelidlessness:
@yanokwa Good news! The failure isn't caused by the logged warnings, rather by the absence of Safari in the test environment. (I'm not sure why it doesn't fail in Node 16.) See:

- [My branch removing Safari](https://github.com/yanokwa/enketo-core/compare/npmjs...eyelidlessness:npmjs)
- [The corresponding CI run passing](https://github.com/eyelidlessness/enketo-core/actions/runs/1419086706)

I don't think we should remove Safari entirely, we still should track the existing failures and eventually get to a state where we can actually include Safari in test runs (at least during dev). For now I think it'd be reasonable to check for `process.env.CI` which will be [`'true'` in GitHub Actions](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables).

Gruntfile.js
```diff
@@ -96,7 +96,7 @@ module.exports = grunt => {
                 browsers: [ 'ChromeHeadlessNoSandbox' ]
             },
             browsers: {
-                browsers: [ 'Chrome', 'Firefox', 'Safari' ]
+                browsers: [ 'Chrome', 'Firefox' ]
```
eyelidlessness, line 99:
```suggestion
                browsers: process.env.CI === 'true' ? [ 'Chrome', 'Firefox' ] : [ 'Chrome', 'Firefox', 'Safari' ]
```

eyelidlessness (changes requested):

eyelidlessness (approved):

PR #832, opened 2021-10-20
Fix: Firefox DatetimepickerExtended tests

eyelidlessness:
- Ensures `originalInputValue` output consistently outputs seconds and milliseconds
- Relaxes test for the value of the widget's backing `datetime-local` input

### DIFF
--- a/test/spec/widget.datetime.spec.js
+++ b/test/spec/widget.datetime.spec.js
@@ -46,7 +46,7 @@ describe( 'datetimepicker widget', () => {
             fakeDateInput.dispatchEvent( new Event( 'change' ) );
 
             // timezone info will be added by engine
-            expect( input.value ).to.equal( '2012-01-01T01:01:00.000' );
+            expect( input.value.endsWith( '-07:00' ) ).to.equal( false );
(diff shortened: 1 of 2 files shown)


test/spec/widget.datetime.spec.js
```diff
@@ -46,7 +46,7 @@ describe( 'datetimepicker widget', () => {
             fakeDateInput.dispatchEvent( new Event( 'change' ) );
 
             // timezone info will be added by engine
-            expect( input.value ).to.equal( '2012-01-01T01:01:00.000' );
+            expect( input.value.endsWith( '-07:00' ) ).to.equal( false );
```
yanokwa, line 49: Seems like `expect( input.value.startsWith( '2012-01-01T01:01' ) ).to.equal( true );` would be clear?

yanokwa (approved):
Standardizing on the output makes sense. So does the test change. In the latter case, I now understand that you are trying to confirm that there is no timezone information in the string. It's not great that we have hard-coded to `-07:00`, but we use that everywhere already.

Issue #835, opened 2021-11-03
loading a record with a meta block in the http://openrosa.org/xforms namespace, fails to properly create and populate a deprecatedID node


MartijnR:
https://github.com/XLSForm/pyxform/issues/105#issuecomment-959796769

* [x] add tests if they do not exist - found one [here](https://github.com/enketo/enketo-core/blob/master/test/spec/form.spec.js#L108)

http://localhost:8005/?xform=meta-namespace.xml

MartijnR:
Thanks! 

I am able to reproduce it with http://localhost:8005/?xform=meta-namespace.xml and ` instanceStr: '<data xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms" id="meta-namespace">\n          <a>ert</a>\n          <b/>\n          <orx:meta>\n              <orx:instanceID>uuid:d5708aa1-2db4-47f6-a98a-bf7962a73905</orx:instanceID>\n              <orx:timeStart>2021-11-03T16:30:30.165-04:00</orx:timeStart>\n              <orx:timeEnd>2021-11-03T16:30:30.166-04:00</orx:timeEnd>\n              <orx:deviceID>localhost:KOGktYtzKR4oqudH</orx:deviceID>\n              <orx:userID/>\n          </orx:meta>\n        </data>'`

This results in:
```xml
 instanceStr: '<data xmlns:jr="http://openrosa.org/javarosa" xmlns:orx="http://openrosa.org/xforms" id="meta-namespace">\n          <a>ert</a>\n          <b/>\n          <orx:meta>\n              <orx:instanceID>uuid:d5708aa1-2db4-47f6-a98a-bf7962a73905</orx:instanceID>\n              <orx:timeStart>2021-11-03T16:30:30.165-04:00</orx:timeStart>\n              <orx:timeEnd>2021-11-03T16:30:30.166-04:00</orx:timeEnd>\n              <orx:deviceID>localhost:KOGktYtzKR4oqudH</orx:deviceID>\n              <orx:userID/>\n          </orx:meta>\n        </data>'
``` 

The `deprecatedID` is created but not populated (1st issue), and not in the correct namespace (2nd issue).

vituslehner:
Hi, I think I'm having a similar issue with submissions that were exported from ODK Aggregate via Briefcase. Those submissions get a `n0:` namespace for meta data which causes submission editing to fail when imported into ODK Central and loaded for editing in Enketo.

[...]

PR #836, opened 2021-11-03
fixed: deprecatedID not added and populated properly if meta block is name-spaced with http://openrosa.org/xforms

MartijnR:
…rg/xforms"-namespaced meta blocks are supported, #835

### DIFF
--- a/src/js/form-model.js
+++ b/src/js/form-model.js
@@ -532,9 +532,13 @@ FormModel.prototype.setInstanceIdAndDeprecatedId = function() {
         // set the instanceID value to empty
         instanceIdEl.textContent = '';
 
+        const namespace = instanceIdEl.namespaceURI;
+
         // add deprecatedID node if necessary
         if ( !deprecatedIdEl ) {
-            deprecatedIdEl = parser.parseFromString( '<deprecatedID/>', 'text/xml' ).documentElement;
+            const nsPrefix = namespace ? this.getNamespacePrefix( namespace ) : '';
+            const nsDeclaration = namespace ? `xmlns:${nsPrefix}="${namespace}"` : '';
+            deprecatedIdEl = parser.parseFromString( `<${nsPrefix ? nsPrefix + ':' : ''}deprecatedID ${nsDeclaration}/>`, 'text/xml' ).documentElement;
             this.xml.adoptNode( deprecatedIdEl );
             metaEl = this.xml.querySelector( '* > meta' );
             metaEl.appendChild( deprecatedIdEl );
--- a/test/spec/formmodel.spec.js
+++ b/test/spec/formmodel.spec.js
@@ -419,8 +419,13 @@ describe( 'DeprecatedID value getter', () => {
         model.init();
         expect( model.deprecatedID ).to.equal( '' );
     } );
-    it( 'returns value of deprecatedID node', () => {
+    it( 'returns value of deprecatedID node if the meta block does not have the "http://openrosa.org/xforms" namespace', () => {
         const model = new Model( '<model><instance><data><meta><instanceID/><deprecatedID>a</deprecatedID></meta></data></instance></model>' );
+        model.init();
+        expect( model.deprecatedID ).to.equal( 'a' );
+    } );
+    it( 'returns value of deprecatedID node if the meta block has the "http://openrosa.org/xforms" namespace', () => {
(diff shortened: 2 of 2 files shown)


test/spec/formmodel.spec.js
```diff
@@ -419,11 +419,16 @@ describe( 'DeprecatedID value getter', () => {
         model.init();
         expect( model.deprecatedID ).to.equal( '' );
     } );
-    it( 'returns value of deprecatedID node', () => {
+    it( 'returns value of deprecatedID node if the meta block does not have the "http://openrosa.org/xforms" namespace', () => {
         const model = new Model( '<model><instance><data><meta><instanceID/><deprecatedID>a</deprecatedID></meta></data></instance></model>' );
         model.init();
         expect( model.deprecatedID ).to.equal( 'a' );
     } );
+    it( 'returns value of deprecatedID node if the meta block has the "http://openrosa.org/xforms" namespace', () => {
```
MartijnR, line 427: Not directly testing the issue but it won't hurt to have this (related) test, I think.

src/js/form-model.js
```diff
@@ -532,9 +532,13 @@ FormModel.prototype.setInstanceIdAndDeprecatedId = function() {
...
         instanceIdEl.textContent = '';
 
+        const namespace = instanceIdEl.namespaceURI;
+
         // add deprecatedID node if necessary
         if ( !deprecatedIdEl ) {
-            deprecatedIdEl = parser.parseFromString( '<deprecatedID/>', 'text/xml' ).documentElement;
+            const nsPrefix = namespace ? this.getNamespacePrefix( namespace ) : '';
+            const nsDeclaration = namespace ? `xmlns:${nsPrefix}="${namespace}"` : '';
+            deprecatedIdEl = parser.parseFromString( `<${nsPrefix ? nsPrefix + ':' : ''}deprecatedID ${nsDeclaration}/>`, 'text/xml' ).documentElement;
```
eyelidlessness, line 541: String concatenation of this always raises my eyebrows a bit. It's probably reasonable enough here, but worth noting for posterity that this is a small risk in cases where special characters are used. Probably more likely for the namespace URI than the prefix, although I was able to cause test failures with both.

test/spec/formmodel.spec.js
```diff
@@ -1201,9 +1206,9 @@ describe( 'merging an instance into the model', () => {
...
             ],
             // namespaces used in both record and model (though now with triple equal namespace declarations..:
-            [ '<a xmlns:cc="http://cc.com"><c>record</c><cc:meta><cc:instanceID>a</cc:instanceID></cc:meta></a>',
-                '<model xmlns:cc="http://cc.com"><instance><a><c/><cc:meta><cc:instanceID/></cc:meta></a></instance></model>',
-                '<model xmlns:cc="http://cc.com"><instance><a xmlns:cc="http://cc.com"><c>record</c><cc:meta xmlns:cc="http://cc.com"><cc:instanceID>a</cc:instanceID></cc:meta></a></instance></model>'
+            [ '<a xmlns:orx="http://openrosa.org/xforms"><c>record</c><orx:meta><orx:instanceID>a</orx:instanceID></orx:meta></a>',
+                '<model xmlns:orx="http://openrosa.org/xforms"><instance><a><c/><orx:meta><orx:instanceID/></orx:meta></a></instance></model>',
+                '<model xmlns:orx="http://openrosa.org/xforms"><instance><a xmlns:orx="http://openrosa.org/xforms"><c>record</c><orx:meta xmlns:orx="http://openrosa.org/xforms"><orx:instanceID>a</orx:instanceID></orx:meta></a></instance></model>'
```
eyelidlessness, line 1211: GitHub won't let me comment inline on the [relevant lines](https://github.com/enketo/enketo-core/pull/836/files#diff-a3816804710c6de44954ba918725ac511d729dd2b01badd447b3d86e2c924951R1237-R1250), so noting here: in the future we should move these lines _into_ the `it` callback. If anything in those lines causes an error, it will terminate the `describe` block without adding the failing test, resulting in a false positive for the test run. (And we should do the same for any other logic executed outside `it`/setup/teardown callbacks. I just happened to notice this case in the course of review.)

eyelidlessness (approved):

test/spec/formmodel.spec.js
MartijnR: The link doesn't seem to work, but I think you're referring to e.g. initializing the model. That's a good point. Thanks! 👍

src/js/form-model.js
MartijnR: Good point. We could create elements and attributes here using createElement etc. Laziness on my part! :)

PR #838, opened 2021-11-05
Improve build speed

yanokwa:
Same as https://github.com/enketo/enketo-transformer/pull/136

### DIFF
--- a/test/spec/form.spec.js
+++ b/test/spec/form.spec.js
@@ -45,8 +45,8 @@ describe( 'Output functionality ', () => {
     form.init();
 
     it( 'tested upon initialization: node random__', () => {
-        const val = form.view.html.querySelector( '[data-value="/random/random__"]' ).textContent;
-        expect( val.length >= 16 && val.length <= 17  ).to.equal( true );
+        const val = Number( form.view.html.querySelector( '[data-value="/random/random__"]' ).textContent );
+        expect( val ).to.be.a( 'number' ).and.to.be.greaterThanOrEqual( 0 ).and.to.be.lessThanOrEqual( 1 );
     } );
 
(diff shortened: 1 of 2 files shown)


test/spec/form.spec.js
```diff
@@ -44,10 +44,10 @@ describe( 'Output functionality ', () => {
...
     form.init();
 
-    it( 'tested upon initialization: node random__', () => {
-        const val = form.view.html.querySelector( '[data-value="/random/random__"]' ).textContent;
-        expect( val.length >= 16 && val.length <= 17  ).to.equal( true );
-    } );
+    // it( 'tested upon initialization: node random__', () => {
+    //     const val = form.view.html.querySelector( '[data-value="/random/random__"]' ).textContent;
+    //     expect( val.length >= 16 && val.length <= 17  ).to.equal( true );
+    // } );
```
eyelidlessness, line 50: This depends on behavior in openrosa-xpath-evaluator, which is this:

```js
random: function() { return XPR.number(parseFloat(Math.random().toFixed(15))); }
```

Which or course may have trailing zeroes, and as `textContent` is of course coerced to a string whose length is being tested here. I think, instead of commenting out the test we can just make it more reliable and more clear what the test is checking, e.g.:

```suggestion
    it( 'tested upon initialization: node random__', () => {
        const val = Number( form.view.html.querySelector( '[data-value="/random/random__"]' ).textContent );

        expect( val ).to.be.a( 'number' ).and.to.be.greaterThanOrEqual( 0 ).and.to.be.lessThanOrEqual( 1 );
    } );
```

I ran this 100,000 times with 100% success.

eyelidlessness (changes requested):

eyelidlessness (approved):

PR #843, opened 2021-11-30
Disable <option>s for readonly <select>s with itemsets

### LINKED ISSUE #842: A readonly `<select>` with an itemset leaves its `<option>`s enabled
**Describe the bug**
A readonly select (minimal or not) will remain operable if its options are provided by an itemset. In Enketo Express this bug may also cause an incorrect value to be shown if a readonly-view sets some options as disabled but any dynamically created options (afterwards) are enabled. A browser will automatically select the first non-disabled option (https://github.com/OpenClinica/enketo-express-oc/issues/520)

**To Reproduce**

#### in enketo-core:
1.  Drop [select-dynamic-readonly.xml.txt](https://github.com/enketo/enketo-core/files/7627019/select-dynamic-readonly.xml.txt) into `test/temp`, change extension to xml 
2. Open a tab in mobile developer view and load http://localhost:8005/?xform=select-dynamic-readonly.xml (in enketo-core)
3. Observe that the value can be changed from milligrams to grams even though the question is readonly

#### in enketo-transformer
1. Drop the above form in test/forms
2. Go to http://localhost:8085/transform/htmlform?xform=http://localhost:8081/select-dynamic-readonly.xml (in enketo-transformer
3. Observe that the option does not have the disabled attribute

**Expected behavior**
Just like static lists of readonly select options (without itemset), the options should be visible but disabled. Note that the readonly attribute, though used in Enketo, actually is not valid on a select (nor on an option).

**Additional context**
An alternative would be to use the `disabled` attribute on readonly selects (and leave options enabled). However, that would prevent users from seeing the options.

Note, until readonly becomes dynamic in Enketo, this issue manifests itself primarily in special readonly views in Enketo Express, e.g. to view an already submitted record.

A fix could likely be created by inspecting the readonly attribute on the parent select in itemset.js and apply the disabled attribute when present.

MartijnR:
Closes #842

Still to do:

* [x] replace enketo-tranformer version after https://github.com/enketo/enketo-transformer/pull/139 has been merged and published
* [x] a FF test is failing (but likely not related to this fix)

### DIFF
--- a/src/js/itemset.js
+++ b/src/js/itemset.js
@@ -97,6 +97,9 @@ export default {
                 input = optionInput.classList.contains( 'ignore' ) ? getSiblingElement( optionInput.closest( '.option-wrapper' ), 'input.rank' ) : optionInput;
             } else if ( list && list.nodeName.toLowerCase() === 'select' ) {
                 input = list;
+                if( input.matches( '[readonly]' ) ){
+                    inputAttributes[ 'disabled' ] = 'disabled';
+                }
             } else if ( list && list.nodeName.toLowerCase() === 'datalist' ) {
                 if ( shared ) {
                     // only the first input, is that okay?
@@ -228,7 +231,7 @@ export default {
                         } else {
                             activeLabelContentNodes = translations[ 0 ].contentNodes;
                         }
-                        optionsFragment.appendChild( that.createOption( activeLabelContentNodes, value ) );
+                        optionsFragment.appendChild( that.createOption( inputAttributes, activeLabelContentNodes, value ) );
                     }
 
                 } );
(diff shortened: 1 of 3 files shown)


eyelidlessness (approved):
I know this is in draft pending release of https://github.com/enketo/enketo-transformer/pull/139, but I can verify it does work as expected. And the change makes sense to me, it took me some time to look through the surrounding code to understand how it works and try to understand what (else) might be affected.

> a FF test is failing (but likely not related to this fix)

I saw several failures on my first run. However I re-ran `grunt transforms` just to be 100% sure, and I no longer see any failures 🙂

lognaturel:
I believe this should have been in 6.0.3 to actually make use of the Transformer update. Perhaps rebase with the goal of releasing as 6.0.4 before releasing Express?

lognaturel:
@MartijnR would you like to do that or should I or @eyelidlessness take over? Would love to close out the week with an Express release if we can!

MartijnR:
If @eyelidlessness can do this that would be great. I won't be able to do it today I think (but next week would be). Thanks!

eyelidlessness:
I'm on it!

lognaturel (approved):

Issue #844, opened 2021-12-10
outputs in labels of radiobutton/checkbox itemsets are not working


MartijnR:
**Describe the bug**
XForm outputs in labels of radiobutton/checkbox itemsets are not working

**To Reproduce**
Steps to reproduce the behavior. If applicable please include the smallest possible XLSForm or XForm (as zip or link).
1. Run [output-dynamic-itemset.xml.txt](https://github.com/enketo/enketo-core/files/7693611/output-dynamic-itemset.xml.txt)
2. Enter 'a' in the first field. Create a new repeat and enter 'b' in that new repeat field.
3. Observe that the dynamic list at the bottom does not show `#a` and `#b`.

**Expected behavior**
The dynamic list at the bottom shows `#a` and `#b`.

**Additional context**

PR #846, opened 2021-12-10
Fix outputs in labels of radiobutton/checkbox itemsets

MartijnR:


src/js/form.js
```diff
@@ -530,7 +530,8 @@ Form.prototype.getRelatedNodes = function( attr, filter, updated ) {
     filter = filter || '';
 
     // The collection of non-repeat inputs, calculations and groups is cached (unchangeable)
-    if ( !this.nonRepeats[ attr ] ) {
+    // We essentially exclude outputs from the cache, because outputs can be added via itemsets (in labels)
+    if ( !this.nonRepeats[ attr ] || filter === '.or-output' ) {
```
MartijnR, line 534: The caching in getRelatedNodes causes an issue for outputs in dynamically update itemset labels. An itemset update doesn't trigger a cache refresh.

It's crude to completely disable caching for outputs though. An alternative would be to only bust the cache if an itemset update deals with options that have outputs.

src/js/itemset.js
```diff
@@ -185,11 +185,11 @@ export default {
                                     const language = label.getAttribute( 'lang' );
                                     const type = label.nodeName;
                                     const src = label.src;
-                                    const text = label.textContent;
+                                    const contentNodes = [ ...label.childNodes ];
```
MartijnR, line 188: Instead of only adding textContent, all nodes will now be added to the itemset (including `<span class="or-output">` nodes)

src/js/itemset.js
```diff
@@ -198,11 +198,11 @@ export default {
                                     // Two falsy values should set active to true.
                                     const active = ( !lang && !that.form.langs.currentLanguage ) || ( lang === that.form.langs.currentLanguage );
 
-                                    return { language: lang, type: 'span', text: label.textContent, active };
+                                    return { language: lang, type: 'span', contentNodes: [ ...label.childNodes ], active };
                                 } );
                                 break;
                             default:
-                                translations = [ { language: '', type: 'span', text: labels && labels.length ? labels[ 0 ].textContent : 'error', active: true } ];
+                                translations = [ { language: '', type: 'span', contentNodes: labels && labels.length ? [ ...labels[ 0 ].childNodes ] : [], active: true } ];
```
eyelidlessness, line 205: This previously populated `'error'` when there were no labels, now appears to populate nothing. Is that intentional? I'm not sure what's expected here.

src/js/itemset.js
eyelidlessness: I was curious about the significance of this change. So I rolled it back (keeping the cache changes), and the difference I see is:

1. The cache changes still partially fix the issue as described in #844, but
2. The labels are not updated until a new repeat is added (also observed on master)

Which in turn made me curious about how each caching change affects behavior, so I:

- Rolled back [both](https://github.com/enketo/enketo-core/pull/846/files#diff-608210f16bb20aa58380a2606ca9a842e85e4634d9e3d03347845848b9bc0f8cR534) [changes](https://github.com/enketo/enketo-core/pull/846/files#diff-608210f16bb20aa58380a2606ca9a842e85e4634d9e3d03347845848b9bc0f8cR550), and indeed I can reproduce #844 again.
- Rolled back only the first change, and the fix still works (I'll add another comment there suggesting it be rolled back, but this was all one thought process so that's why I'm noting it here).
- Rolled back only the second change, and again only see labels update when adding a new repeat.

eyelidlessness (changes requested):

src/js/form.js
```diff
@@ -544,8 +545,9 @@ Form.prototype.getRelatedNodes = function( attr, filter, updated ) {
     }
 
     // If a new repeat was created, update the cached collection of all form controls with that attribute
-    // If a repeat was deleted ( update.repeatPath && !updated.cloned), rebuild cache
-    if ( !this.all[ attr ] || ( updated.repeatPath && !updated.cloned ) ) {
+    // If a repeat was deleted ( update.repeatPath && !updated.cloned), rebuild cache.
+    // Exclude outputs from the cache, because outputs can be added via itemsets (in labels).
+    if ( !this.all[ attr ] || ( updated.repeatPath && !updated.cloned ) || filter === '.or-output' ) {
```
eyelidlessness, line 550: More of a general question than anything particular to this change/review: it's concerning to me that responsibilities for discrete elements/widgets like this are spread out across so many parts of the codebase.

I wonder if we could gradually move toward a model where those discrete things provide an interface for these kinds of conditionals/branches and the form could delegate to those? I'm not sure how that would work with this specifically, but it's something I've been thinking about generally.

src/js/itemset.js
```diff
@@ -318,12 +318,13 @@ export default {
      * Creates a HTML option element
      *
      * @param {string} label - option label
```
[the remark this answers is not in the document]
MartijnR: Ok, sure!

src/js/itemset.js
MartijnR: Ah nice find! Indeed, I misread the code and the second cache change is not required!

src/js/form.js
MartijnR: Yes, this is concerning to me too. It starts off elegant and then slowly grows into a monster. This is especially the case with performance tweaks like this (no caching at all would create much cleaner code). Definitely could use some redesign.

src/js/form.js
MartijnR: It's the second caching change that can be rolled back, that you discovered, I believe. Not this one.

The alternative implementation would create complexity. Either directly clearing the cache in itemset.js (which seems wrong), or use a new event and listener, I think.

src/js/itemset.js
MartijnR: It's intentional. I think it makes more sense to show an empty label if those nodes are missing or falsy.

eyelidlessness (approved):

PR #866, opened 2022-03-01
Code style/formatting: Prettier + Airbnb ESLint preset

eyelidlessness:
The purpose of this change is to move toward a more idiomatic code style, and automate as much of it as possible. What's in the change:

- Adds automatic formatting provided by [Prettier](https://prettier.io/)
- Consolidates all ESLint configs into one file via `overrides`
  - Overrides files run in a Node environment to specify that in ESLint config (removing arbitrary global declarations there and across several modules)
- Extends [Airbnb's ESLint config](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb), a commonly used, opinionated set of rules
  - Overrides some of those rules to be warnings where they're not automatically fixable; there are _a lot of warnings_, the goal is to fix them wherever we touch code with warnings
  - Manually fixes a very small portion of those warnings where it was trivial and obviously safe (`no-redeclare` for Node globals, a couple of optional function params before required params which never receive `undefined`/never hit the fallback value)
- Extends Prettier's ESLint config to enable checking Prettier rules (e.g. while editing in VSCode and in CI)
- Applies automatic Prettier and ESLint formatting on save in VSCode

Currently opened as draft while I continue manual testing.


eyelidlessness:
*Sigh, once again*. It appears tests depend on imports which are generated by transforming forms, and ignored by git. I'll address this tomorrow.

lognaturel (approved):
Scrubbed through quickly to confirm it matches what you did in Express and doesn’t present obvious risk.

Issue #849, opened 2022-01-05
Only use relevant values in calculations (e.g., like Collect)


yanokwa:
In Collect, when a question is not relevant, its value is treated as blank for all expressions it is used in, including relevance. In Enketo, even irrelevant values are used in calculations. This difference between Collect and Enketo is unexpected and we think Enketo's behavior should change because it's the least widely deployed. 

This change will most certainly break forms in folks using Enketo outside of ODK core tools and so we do need to be cautious about how this is communicated and rolled out.

More discussion (including references to the vagueness of the XForms spec on this issue) are at https://forum.getodk.org/t/enketo-v3-uses-non-relevant-values-in-calculations-whereas-collect-does-not/35567/

MartijnR:
> its value is treated as blank for all expressions it is used in

Does this mean that the value is set to empty in the model as soon as it becomes non-relevant?

Will try to find Enketo's reasoning fwiw. I believe it was done to not loose user-entered values if users accidentally make a group with entered data non-relevant (and because the XForms spec talks about "pruning" non-relevant values before submission somewhere if I remember correctly).

There is already a mechanism (search for `emptyNonRelevant` in [calculate.js](https://github.com/enketo/enketo-core/blob/master/src/js/calculate.js) to make this change. I think OpenClinica always clears non-user-entered values in questions with calculations when they become non-relevant, and it uses that `emptyNonRelevant` parameter to provide this customization.

related bug (in repeats): https://github.com/enketo/enketo-core/issues/830

lognaturel:
> Does this mean that the value is set to empty in the model as soon as it becomes non-relevant?

No! It's part of the evaluation logic to check relevance and treat values that are non-relevant as blank. So if a value becomes relevant again, it does still have the user-input value. The pruning happens when a user leaves the current form editing session.

MartijnR:
Ah, in that case ignore what I wrote above about the existing mechanism.

```xml
<data> 
   <a>2</a>
   <b>3</b>
</data>
```

So just to be sure, if a calculation `/data/a + /data/b` is sent to the XPath evaluator with the above model, and `/data/a` is non-relevant, the result would be `3`?

lognaturel:
> concat(/data/a, data/b) would be "b"

I think you mean `3` and yes, that's right.

> /data/a + /data/b is sent to the XPath evaluator with the above model, and /data/a is non-relevant but /data/b is relevant, the result would be NaN

[...]

PR #870, opened 2022-03-07
Optionally clear non-relevant values/remove them from calculations

eyelidlessness:
Addresses #849.

When `excludeNonRelevant` is `true` in config.js:

- Model nodes which become non-relevant have their values stored as an `non-relevant-value` attribute rather than as `textContent`, allowing values to continue to be updated but excluded from calculations
- Model nodes which become relevant have their values restored to `textContent`, restoring their use in calculations

This behavior is optional and off by default, to avoid unnecessary breaking changes. Dependent projects such as Enketo Express will need to opt into this by overriding the default configuration.


MartijnR:
Out of curiosity I glanced quickly at the code. That is a pretty nifty approach to this issue!! 👍  

I cannot (yet) quite wrap my head around how restoring cleared values will work with nested relevant conditions (when these become relevant in a particular order).

lognaturel:
It was too good to be true! @eyelidlessness did an interactive review and came up with a couple of scenarios that would fail. He's writing up tests and exploring a variation on the theme.

MartijnR:
> @MartijnR when you mentioned serious performance degradation at https://github.com/enketo/enketo-core/issues/849#issuecomment-1011445891, is this what you were concerned about or worse?

No, I was referring to having to pass a pruned version of the XML model to the evaluator every time an expression gets evaluated which was assuming we'd leave the values in place. But your approach is avoiding that issue.

eyelidlessness:
A few things I've learned with the form being used in QA:

- This does significantly degrade performance, probably unacceptably so
- I'm going to need to add caching to core's transformations to be productive testing with said form, because that's slow too

So now I have a form I can use to really assess some performance issues, yay! That said, even running an XML formatter on the file in VSCode takes several seconds, so it's probably going to be slowish in some respects no matter what our performance characteristics.

lognaturel:
@eyelidlessness Now that a bunch of what was here has been spun out and merged, I can't wait to see this distilled to just the relevance change!

config.js
```diff
@@ -11,7 +11,11 @@ export default /** @type {const} */ ({
         computeAsync: window.location.search.includes('&computeAsync'),
     },
 
-    forceClearNonRelevant: window.location.search.includes('&clearNonRelevant'),
+    /**
+     * When set to `true`, non-relevant values will be treated as blank. This behavior
+     * is largely consistent with JavaRosa.
```
lognaturel, line 16: largely? That sounds pessimistic? 😄

eyelidlessness: If I recall, that wording is a reference to [this](https://github.com/enketo/enketo-core/pull/870/files#diff-1639f2fb93ca42c1a35786766efa8b81527f63010903193dcf6afd8cf685334fR391-R392). 🙃

lognaturel: Can you try that link again? I think it might have been on a commit that's gone.

test/spec/repeat.spec.js
```diff
@@ -401,19 +410,24 @@ describe('repeat functionality', () => {
                 <rep>
                     <txt/>
                     <num>5</num>
+                    <is-num-relevant>yes</is-num-relevant>
```
lognaturel, line 413: Were these perhaps changes that were never used? Or just wanted to exactly match https://github.com/enketo/enketo-core/pull/870/files#diff-f2896b2fc4580368e34c58eec902cffe5d8e202a73f906034a965b8be7c6ebaaR875?

test/spec/repeat.spec.js
```diff
@@ -710,7 +725,581 @@ describe('repeat functionality', () => {
             form.init();
             // Issue where a calculation inside a repeat is cached before the repeats are initialized (which removes the first repeat, before adding it)
             // This results in two cached calculations (for the same node) of which one no longer exists.
-            expect(form.getRelatedNodes('data-calculate').length).to.equal(1);
+            expect(
+                form.getRelatedNodes('data-calculate', '[name="/data/rg/row"]')
```
lognaturel, line 729: What is this change for?

test/spec/calculate.spec.js
```diff
@@ -146,9 +214,662 @@ describe('calculate functionality', () => {
...
+            didHaveFever.dispatchEvent(events.Change());
+
+            timers.runAll();
+
+            expect(feverDurationDayUnitModelNode.textContent).to.equal('days');
+            expect(feverDurationDaysModelNode.textContent).to.equal('36');
+        }, 1000000);
+
+        // This test, as with the larger one above, should be moved to a relevance-specific suite.
+        // But I am le tired.
```
lognaturel, line 810: Still le tired?

An alternative would be to pull out tests on this form into their own world. In JR it has acted as a general smoketest because there's so much going on in it.

lognaturel (commented):
One thing I'm not quickly finding -- tests around grandparent relevance. That is, a node with no relevance expression inside of a node with a relevance expression itself inside of a node with a relevance expression. Toggle the outer relevance and ensure the innermost relevance updates accordingly.

config.js
eyelidlessness: I think it's just GitHub failing to expand the diff in the link? [Here it is](https://github.com/enketo/enketo-core/blob/e6c24bd9ed216115a84b4e5c1b9944055540a083/test/spec/calculate.spec.js#L391).

test/spec/repeat.spec.js
eyelidlessness: Another leftover from changes extracted into new forms 🙃

test/spec/repeat.spec.js
eyelidlessness: Missed when extracting changes into new forms.

src/js/relevant.js
```diff
@@ -138,10 +213,10 @@ export default {
                     // The path is stripped of the last nodeName to record the context.
                     // This might be dangerous, but until we find a bug, it helps in those forms where one group contains
                     // many sibling questions that each have the same relevant.
-                    cacheIndex = `${p.relevant}__${p.path.substring(
-                        0,
-                        p.path.lastIndexOf('/')
-                    )}__${p.ind}`;
+                    cacheIndex = `${p.relevant}__${
+                        repeatPath ??
+                        p.path.substring(0, p.path.lastIndexOf('/'))
```
eyelidlessness, line 218:
```suggestion
                        p.path
```

src/js/form.js
```diff
@@ -356,6 +369,12 @@ Form.prototype.init = function () {
             });
 
             this.all = {};
+
+            this.repeatPathPrefixes = Array.from(
+                this.view.html.querySelectorAll('.or-repeat-info')
+            )
+                .map((element) => `${element.dataset.name}/`)
+                .sort((a, b) => b.length - a.length);
```
lognaturel, line 377: Repeat path prefixes sorted with the deepest ones first.

src/js/nodeset.js
```diff
@@ -119,12 +120,19 @@ Nodeset.prototype.setVal = function (newVals, xmlDataType) {
...
+    const strVal = String(newVal);
     const targets = this.getElements();
 
-    if (targets.length === 1 && newVal.toString() !== curVal.toString()) {
+    if (targets.length === 1 && strVal !== curVal.toString()) {
         const target = targets[0];
-        // first change the value so that it can be evaluated in XPath (validated)
-        target.textContent = newVal.toString();
+        // First change the value so that it can be evaluated in XPath (validated).
+        if (isNodeRelevant(target)) {
```
lognaturel, line 130: `isNodeRelevant` is always true if the relevance behavior feature flag is off.

src/js/relevant.js
```diff
@@ -226,19 +308,175 @@ export default {
...
+     * @property {number} repeatIndex
+     * @property {string} repeatPath
+     */
+
+    /**
+     * @param {HTMLElement} branchNode
+     * @param {string} path
+     * @param {ToggleNonRelevantModleNodesOptions} options
+     */
+    toggleNonRelevantModelNodes(branchNode, path, options) {
```
lognaturel, line 333: This is the bulk of it, challenge was balancing correctness and performance

TODO: H come back

src/js/relevant.js
```diff
@@ -65,29 +108,50 @@ export default {
...
-                    .filter(
-                        (group) =>
-                            getChild(group, '.or-repeat') ||
-                            !getChild(group, '.or-repeat-info')
-                    );
-                // If the parent doesn't exist in the DOM it means there is a repeat ancestor and there are no instances of that repeat.
-                // Hence that relevant does not need to be evaluated (and would fail otherwise because the context doesn't exist).
-                if (parentGroups.length === 0) {
-                    return;
+            const { repeatIndex } = options;
```
lognaturel, line 111: Performance changes:
- Don't do repeat logic if there are no repeats
- Look in the caches before doing any logic

Discovered as part of making things correct.

lognaturel (commented):
Add a test with nested repeat and in the innermost repeat, there are two nodes, the relevance of the second references the first one relatively.

lognaturel (approved):

lognaturel:
I'm feeling ready for this to be merged!

I focused the most energy on making sure I understand changes that are not feature-flagged and that they're low risk. We plan on releasing Express with the feature flag and continuing to do quality assurance and solicit broader feedback.

PR #871, opened 2022-03-10
[fix] escape querySelector variables to avoid error.

nanzhang01:
**Description of change**
 Escape css selector because the parameter may contains spatial characters.

**Related issue** 
#825


eyelidlessness (commented):
Thanks for the PR @nanzhang01! I have also been wanting to address this, it's concerned me whenever I see these sorts of concatenations... but I've held off because I didn't know `CSS.escape` was a thing. TIL!

There are some merge conflicts due to code style changes merged in #866. I'd also like to consider:

- This could probably be generalized, so...
- ... we can also add test coverage without testing every case where it's used

I'm happy to pick this up and make those changes, but it may take me a bit to get to that. If you'd like to address them in the meantime, I'd be happy to take another look. 🙂

lognaturel:
> add test coverage without testing every case where it's used

That would be great. Can we get this in as it is and file an issue?

eyelidlessness (approved):
Thank you! I'll follow up with an issue to track generalizing and testing.

PR #873, opened 2022-03-10
Fix: improve error messages for missing external instances

eyelidlessness:
Part of enketo/enketo-express#363

### DIFF
--- a/test/spec/form.spec.js
+++ b/test/spec/form.spec.js
@@ -1724,6 +1724,77 @@ describe('Form.prototype', () => {
     });
 });
 
+describe('Missing external instances', () => {
+    it('does not produce extraneous/unrelated errors', () => {
+        const options = { full: true };
+        const session = {
+            deviceid: 'a',
+            username: 'b',
+            email: 'c',
+            phonenumber: 'd',
+            simserial: 'e',
+            subscriberid: 'f',
+        };
+        const externalInstances = [undefined, null];
+        const form = loadForm(
+            'missing-external-instance.xml',
+            undefined,
+            options,
+            session,
+            externalInstances
+        );
+        const loadErrors = form.init();
+
+        loadErrors.forEach((error) => {
+            expect(error).not.to.include(
+                'FormLogicError: instance "__session" does not exist in model'
+            );
+            expect(error).not.to.include('Error trying to parse XML model');
+        });
+
+        expect(loadErrors).to.deep.equal([
+            "Can't find brands.csv.",
+            "Can't find products.csv.",
+        ]);
+    });
+
+    it('does not produce extraneous/unrelated errors', () => {
+        const options = { full: true };
+        const session = {
+            deviceid: 'a',
+            username: 'b',
+            email: 'c',
+            phonenumber: 'd',
+            simserial: 'e',
+            subscriberid: 'f',
+        };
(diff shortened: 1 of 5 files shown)


test/spec/form.spec.js
```diff
@@ -1724,6 +1724,77 @@ describe('Form.prototype', () => {
...
+            expect(error).not.to.include('Error trying to parse XML model');
+        });
+
+        expect(loadErrors).to.deep.equal([
+            "Can't find brands.csv.",
+            "Can't find products.csv.",
+        ]);
+    });
+
+    it('does not produce extraneous/unrelated errors', () => {
```
lognaturel, line 1761: I think this name is supposed to be different? Is it intended to show the empty attachment error? I’m surprised that the error is the same as above.

test/spec/form.spec.js
```diff
@@ -1724,6 +1724,77 @@ describe('Form.prototype', () => {
     });
 });
 
+describe('Missing external instances', () => {
+    it('does not produce extraneous/unrelated errors', () => {
+        const options = { full: true };
+        const session = {
```
lognaturel, line 1730: This could just be an empty session, right? As-is it looks like the session is meaningful.

test/spec/form.spec.js
lognaturel: Ahh, got it. So the first test has `undefined` and `null`. And this one has defined sources but no file. To me it would make more sense to either have all three in one form or three different tests. I think having all that `session` stuff also made me miss the difference between the two.

I think adding `when the value is missing` and `when there is a value but the data for the external instance is missing` would really help.

I don't know what a table-style test is. 😬

eyelidlessness: > I don't know what a table-style test is. 😬

Ah sorry, I’m not even sure what the appropriate jargon for it is! Parameterized tests? I’m referring to looping over a set of fixtures to use in the test. Commonly used as `it.each` in other JS test frameworks, often as in a standard `[...].forEach`. I tend to avoid it unless it’s a sure thing that it won’t get branchy, because test helpers are hard to untangle once they exist.

I agree the changes you suggested make sense. I’ll get that pushed up tomorrow.

test/spec/form.spec.js
eyelidlessness: The session was actually meaningful for one of the errors this change eliminated. It’s been a while, so I can’t recall if (or why exactly) I had to populate the whole thing… it’s possible a smaller subset would work and I can check that tomorrow too. But it definitely didn’t reproduce the error with an empty session.

test/spec/form.spec.js
lognaturel: Yes, yes, I'm with you all around!

test/spec/form.spec.js
```diff
@@ -1724,6 +1724,65 @@ describe('Form.prototype', () => {
...
+        ];
+        const form = loadForm(
+            'missing-external-instance.xml',
+            undefined,
+            options,
+            session,
+            externalInstances
+        );
+        const loadErrors = form.init();
+        console.log('loadErrors', loadErrors);
```
lognaturel, line 1770: Maybe remove logs before merge? Other than that looks great.

lognaturel (approved):

PR #878, opened 2022-04-07
Update `datepicker-extended` widget to properly unpack data from paste event

### LINKED ISSUE #877: Pasting value into datepicker-extended widget causes error
**Describe the bug**
When using the `datepicker-extended` widget you can paste text from the clipboard into the into the widget's text box.  However, doing so causes an error to be printed to the browser console and, if the pasted value is an invalid value, the value is not immediately cleared from the widget.

Once the widget is closed, the pasted value seems to be properly handled (valid values are set and invalid values cause the value to be cleared). But, until the widget is closed the experience can be weird.  E.g. see the screenshot below for what happens when just pasting in `1` (remember this is the `datepicker-extended` widget in the screenshot!):

image

This is the error that gets printed:
```
Uncaught TypeError: Cannot read properties of null (reading 'length')
```

**To Reproduce**

1. Open a form with a `date` question that has the `year` appearance
    1. (The error happens regardless of the appearance, but you need the `year` appearance to recreate the screenshot above...)
2. Click on the date question to open the widget
3. Paste in the value `1`
4. See the error in the console logs

**Expected behavior**

The error is coming from this line: https://github.com/enketo/enketo-core/blob/51c5c2f494f1515a67355543b435f6aaa4b151b4/src/widget/date/datepicker-extended.js#L94

Instead of throwing an error, this code should be properly clearing out any invalid values that are pasted into the widget.

-   OS: Linux (Mint 20.3)
-   Browser: Chromium 99.0.4844.84

jkuester:
When pasting values into the `datepicker-extended` widget, the following error is printed to the browser console:

```
Uncaught TypeError: Cannot read properties of null (reading 'length')
```

This is because the widget code is not getting the `originalEvent` from the jQuery event wrapper before unpacking the paste data.

(It is worth noting that the only other place the `getPasteData` util function is being used, in [`mask.js`](https://github.com/enketo/enketo-core/blob/51c5c2f494f1515a67355543b435f6aaa4b151b4/src/js/mask.js#L54), appears to work fine since it is passing the normal browser event (and not the jQuery event)).

Closes #877

### DIFF
--- a/src/widget/date/datepicker-extended.js
+++ b/src/widget/date/datepicker-extended.js
@@ -90,7 +90,9 @@ class DatepickerExtended extends Widget {
         $fakeDateI.on('change paste', (e) => {
             let convertedValue = '';
             let value =
-                e.type === 'paste' ? getPasteData(e.originalEvent ?? e) : this.value;
+                e.type === 'paste'
+                    ? getPasteData(e.originalEvent ?? e)
+                    : this.value;
 
             if (value.length > 0) {
                 // Note: types.date.convert considers numbers to be a number of days since the epoch
--- a/src/widget/date/datepicker-extended.js
+++ b/src/widget/date/datepicker-extended.js
@@ -89,7 +89,7 @@ class DatepickerExtended extends Widget {
 
         $fakeDateI.on('change paste', (e) => {
             let convertedValue = '';
-            let value = e.type === 'paste' ? getPasteData(e) : this.value;
+            let value = e.type === 'paste' ? getPasteData(e.originalEvent) : this.value;
 
             if (value.length > 0) {
                 // Note: types.date.convert considers numbers to be a number of days since the epoch
--- a/src/widget/date/datepicker-extended.js
+++ b/src/widget/date/datepicker-extended.js
@@ -89,7 +89,8 @@ class DatepickerExtended extends Widget {
 
         $fakeDateI.on('change paste', (e) => {
             let convertedValue = '';
-            let value = e.type === 'paste' ? getPasteData(e.originalEvent) : this.value;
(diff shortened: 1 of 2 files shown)


src/widget/date/datepicker-extended.js
```diff
@@ -89,7 +89,8 @@ class DatepickerExtended extends Widget {
 
         $fakeDateI.on('change paste', (e) => {
             let convertedValue = '';
-            let value = e.type === 'paste' ? getPasteData(e) : this.value;
+            let value =
+                e.type === 'paste' ? getPasteData(e.originalEvent) : this.value;
```
eyelidlessness, line 93: We've had some discussion about generally moving away from jQuery. I've recently spotted some code where jQuery code has been partially migrated to regular DOM code, but still partially using jQuery APIs, causing the code not to run.

The tests would _hopefully_ catch that, but as a bit of extra precaution in case they didn't, I'd like to consider falling back to the event received in the callback like:

```suggestion
                e.type === 'paste' ? getPasteData(e.originalEvent ?? e) : this.value;
```

eyelidlessness (changes requested):
Thanks for this PR, and the added test coverage! I requested one small change inline, but otherwise this looks good.

src/widget/date/datepicker-extended.js
jkuester: Nice!  +1 on moving away from jQuery!  (Also TIL about [`??`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) so thanks for that!)

eyelidlessness (approved):

PR #881, opened 2022-04-19
Optionally recompute the evaluation cascade asynchronously

eyelidlessness:
### Async evaluation cascade

Currently under the feature flag `config.experimentalOptimizations.computeAsync`, this allows the evaluation cascade to be performed asynchronously. This improves UI responsiveness for expensive calculations, especially when chained. Currently the feature flag can be enabled dynamically by appending `&computeAsync` to the URL when running the validation app.

This is achieved using [`requestIdleCallback`](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback)[^1], which calls the callback on when the event loop is idle or after `timeout` milliseconds, whichever comes first. The default `timeout` of `16` is approximately the number of milliseconds which would achieve 60 frames per second. This was chosen over `requestAnimationFrame` because the latter is higher priority, calling the callback _before_ the next tick of the event loop.

[^1]: Where available. In browsers which don't currently support `requestIdleCallback` (currently Safari, unless enabled with a flag), it will fall back to `setTimeout` with no delay.

### Risks of async

While I don't believe this specific change introduces a possibility of a race condition, it's _possible_ I have missed a case where it would. As MDN notes:

> Functions are generally called in first-in-first-out order; however, callbacks which have a `timeout` specified may be called out-of-order if necessary in order to run them before the timeout elapses.

In my experience, this has not been the case. I expect that's because callbacks requested with the `timeout` value are effectively placed in a priority queue. When `timeout` is always the same value, the priority should always be the same. But there's a possibility I've misunderstood this and the evaluation cascade order may be unstable with this setting.

My goal in calling this out is:

- To have my understanding of this risk validated/scrutinized in review
- If we determine the risk is as low as I believe, consider whether to remove the feature flag and make this the default behavior
- Otherwise, consider ways we can have users test/validate the behavior

### Validation of perceived performance with async

This change also (likely temporarily, pending #870) reintroduces an option `config.forceClearNonRelevant` to clear non-relevant values. This flag can also be enabled by appending `&clearNonRelevant` to the validation app URL. This differs from #870 in that it will not restore values which have been cleared on questions which become relevant again. This setting was introduced to simulate similar performance issues found in #870 on the large WHO form (also included in this PR).

The perceived performance difference can be observed by comparing:

1. Open the large WHO form
2. Click Go to End
3. Answer "Yes" to "Did the respondent give consent?"
4. Switch the answer to "No"

What you should see:

- With both flags off, there's a brief delay on step 3 before the "Yes" radio appears checked. Step 4, and any repetition of steps 3-4 should feel instantaneous.
- With only `&clearNonRelevant`, both steps 3 and 4 have a noticeably longer delay, even when repeated.
- With both flags enabled, steps 3 and 4 should feel instantaneous or nearly so each time.

### Other changes and details

In the course of validating these changes against a few other forms [found here](https://forum.getodk.org/t/share-your-biggest-or-slowest-form-and-win-a-badge/8984), I noticed that the SOAR form also included in this PR was quite slow to load.

Initially I considered applying the same async approach to `Form#init`, but backed out of this because:

- It would change the method's signature. Although this could be isolated with the feature flag, that would not be possible should this ever become the default approach.
- I couldn't find an effective way, with the SOAR form, to improve _perceived_ initial load time (roughly analogous to [First Contentful Paint](https://developer.mozilla.org/en-US/docs/Glossary/First_contentful_paint)) without introducing quite a lot of delayed repaints. This felt like a worse user experience than simply accepting the full loading time delay.

That said, I discovered on that form that the largest performance impact on load was recalculating for each repeat as its template was added to the form. I was able to improve the _actual_ load time significantly (from about 8-10 seconds on my computer to about 2-3 seconds) by performing all of those calculations at once after all repeats are added.

### Risks of init repeat calculation changes

I refined this portion of the change several times, with each earlier approach failing certain tests. My hope is that with those tests failing, then passing, there is enough coverage that any other behavior change I might have missed would be caught. But my confidence in that is slightly lower than my confidence in the safety of the async change, so I feel it's important to call that out. Repeat behavior is tested _extensively_, of course, with many edge cases considered. But I'm hoping review will validate that this change is desired and safe.

I didn't place this change behind a feature flag, mainly because I forgot to do that in haste of opening this PR. I'd be happy to make that change if desired.

### DIFF
--- a/src/js/form.js
+++ b/src/js/form.js
@@ -98,19 +98,25 @@ Form.prototype = {
             this.validationUpdate,
         ];
 
-        let evaluationCascade;
+        const { evaluationCascadeAdditions } = this;
+
+        if (evaluationCascadeAdditions.length > 0) {
+            baseEvaluationCascade.push(() => {
+                for (const fn of evaluationCascadeAdditions) {
+                    fn();
+                }
+            });
+        }
 
         if (config.experimentalOptimizations.computeAsync) {
-            evaluationCascade = baseEvaluationCascade.map(
+            return baseEvaluationCascade.map(
                 (fn) =>
                     (...args) =>
                         callOnIdle(() => fn(...args))
             );
-        } else {
-            evaluationCascade = baseEvaluationCascade;
         }
 
-        return evaluationCascade.concat(this.evaluationCascadeAdditions);
+        return baseEvaluationCascade;
     },
     /**
      * @type {string}
--- a/src/js/form.js
+++ b/src/js/form.js
@@ -33,6 +33,7 @@ import FormLogicError from './form-logic-error';
 import events from './event';
 import './plugins';
 import './extend';
+import { callOnIdle } from './timers';
 
 /**
  * Class: Form
@@ -86,7 +87,7 @@ Form.prototype = {
      * @type {Array}
      */
     get evaluationCascade() {
-        return [
+        const baseEvaluationCascade = [
             this.calc.update.bind(this.calc),
             this.repeats.countUpdate.bind(this.repeats),
             this.relevant.update.bind(this.relevant),
@@ -95,7 +96,21 @@ Form.prototype = {
             this.required.update.bind(this.required),
             this.readonly.update.bind(this.readonly),
             this.validationUpdate,
-        ].concat(this.evaluationCascadeAdditions);
+        ];
+
+        let evaluationCascade;
+
+        if (config.experimentalOptimizations.computeAsync) {
+            evaluationCascade = baseEvaluationCascade.map(
+                (fn) =>
+                    (...args) =>
+                        callOnIdle(() => fn(...args))
+            );
+        } else {
+            evaluationCascade = baseEvaluationCascade;
+        }
+
+        return evaluationCascade.concat(this.evaluationCascadeAdditions);
     },
     /**
      * @type {string}
--- a/config.js
+++ b/config.js
@@ -1,4 +1,18 @@
-export default {
+export default /** @type {const} */ ({
+    experimentalOptimizations: {
+        /**
+         * When set to `true`, recomputations of the evaluation cascade will be performed
+         * asynchronously to reduce time spent blocking UI updates. This may be improve
+         * perceived performance on large forms with complex chained computations. These
+         * computations are technically delayed and will perform more slowly, but their
+         * corresponding UI updates will render more quickly as each step in the chain of
+         * computations completes.
+         */
+        computeAsync: window.location.search.includes('&computeAsync'),
+    },
+
+    forceClearNonRelevant: window.location.search.includes('&clearNonRelevant'),
+
     maps: [
         {
             tiles: ['https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png'],
@@ -17,4 +31,4 @@ export default {
     validatePage: true,
     swipePage: true,
     textMaxChars: 2000,
-};
+});
--- a/src/js/relevant.js
+++ b/src/js/relevant.js
@@ -4,6 +4,7 @@
  * @description Updates branches
  */
 
+import config from 'enketo/config';
 import events from './event';
 import { closestAncestorUntil, getChild, getChildren } from './dom-utils';
 
@@ -12,7 +13,7 @@ export default {
      * @param {UpdatedDataNodes} [updated] - The object containing info on updated data nodes.
      * @param {boolean} forceClearNonRelevant -  whether to empty the values of non-relevant nodes
      */
-    update(updated, forceClearNonRelevant = false) {
+    update(updated, forceClearNonRelevant = config.forceClearNonRelevant) {
         if (!this.form) {
             throw new Error(
                 'Branch module not correctly instantiated with form property.'
@@ -29,7 +30,7 @@ export default {
      * @param {Array<Element>} nodes - Nodes to update
      * @param {boolean} forceClearNonRelevant - whether to empty the values of non-relevant nodes
      */
-    updateNodes(nodes, forceClearNonRelevant = false) {
+    updateNodes(nodes, forceClearNonRelevant = config.forceClearNonRelevant) {
         let branchChange = false;
         const relevantCache = {};
         const alreadyCovered = [];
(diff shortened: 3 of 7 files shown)


src/js/form.js
```diff
@@ -313,26 +328,32 @@ Form.prototype.init = function () {
 
         // after radio button data-name setting (now done in XLST)
         // Set temporary event handler to ensure calculations in newly added repeats are run for the first time
-        const tempHandlerAddRepeat = (event) => this.calc.update(event.detail);
-        const tempHandlerRemoveRepeat = () => (this.all = {});
```
MartijnR, line 317: I don't immediately see if your changes eliminated the need for this remove handler, or that there wasn't a need in the first place. I'm guessing this removed remove handler was for a form with a default repeat (or several defaults which is possible in XForm, though not in XLSForm) but with a repeat-count of 0. I just wanted to flag this in case there isn't a test for this.

Gruntfile.js
```diff
@@ -169,44 +177,90 @@ module.exports = (grunt) => {
 
     grunt.loadNpmTasks('grunt-sass');
 
-    grunt.registerTask('transforms', 'Creating forms.js', function () {
-        const forms = {};
-        const done = this.async();
-        const jsonStringify = require('json-pretty');
```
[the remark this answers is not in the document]
MartijnR: This looks great. I think it should stay in its own commit (not squashed). I think it should be in its own PR actually (unless I am missing the link).

src/js/form.js
```diff
@@ -313,26 +328,32 @@ Form.prototype.init = function () {
...
         );
-        this.view.html.addEventListener(
-            events.RemoveRepeat().type,
-            tempHandlerRemoveRepeat
-        );
         this.repeatsInitialized = true;
         this.repeats.init();
+
+        // Recalculate after repeats are initialized. Previously this was performed in
+        // `tempHandlerAddRepeat`, but recalculating them all at once is significantly faster.
```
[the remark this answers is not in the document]
MartijnR: I would also separate this mostly because I think the asynchronous evaluation might turn out to be a very significant change and should be in its own commit without distractions.

MartijnR (commented):
This is very intriguing!!

I only looked at the code so far, and would love to see this in action (and comment on the risks as you asked). I aim to do so tomorrow. However, I may not manage to do that, and in that case it would take another week at least. So I just wanted to share these initial comments already.

MartijnR:
>  the evaluation cascade order may be unstable with this setting

I am wondering if a slow XPath expression could throw off the order by completing after a fast subsequent expression. I'm guessing e.g. `count(instance('d')//a)` could be slow if the instance is large.

app.js
```diff
@@ -27,12 +27,36 @@ if (getURLParameter('touch') === 'true') {
...
+                    };
+                }
+            }
+
+            if (survey == null) {
+                const response = await fetch(transformerUrl);
+
+                survey = await response.json();
+            }
+
```
MartijnR, line 59: I am puzzled how this change appears to have made the WHO form so much faster to load (even though `initializeForm` is slightly slower).

src/js/form.js
```diff
@@ -95,7 +96,21 @@ Form.prototype = {
...
+            evaluationCascade = baseEvaluationCascade.map(
+                (fn) =>
+                    (...args) =>
+                        callOnIdle(() => fn(...args))
+            );
+        } else {
+            evaluationCascade = baseEvaluationCascade;
+        }
+
+        return evaluationCascade.concat(this.evaluationCascadeAdditions);
```
MartijnR, line 113: would the (async) `evaluationCascadeAdditions` now run before the (sync) `baseEvalationCascade`? Should this not be concatenated before the if statement?

MartijnR:
> But my confidence in that is slightly lower than my confidence in the safety of the async change

Fwiw, I cannot think of a problem with the repeat initialization change. Perhaps a subtle undiscovered (but fixable) bug in the caching by `getRelatedNodes`. But I think that part is ready to merge and does not need a flag.

src/js/form.js
eyelidlessness: I kept what I believe to be the same logic, it just also moved to after repeat initialization [here](https://github.com/enketo/enketo-core/pull/881/files/5e9c82fc29cc78ca1c9b79423d5a24b94a669800#diff-608210f16bb20aa58380a2606ca9a842e85e4634d9e3d03347845848b9bc0f8cR351). I will double check, but I'm reasonably certain that a test with a repeat count of 0 did fail without that line.

eyelidlessness:
> > the evaluation cascade order may be unstable with this setting
> 
> I am wondering if a slow XPath expression could throw off the order by completing after a fast subsequent expression. I'm guessing e.g. `count(instance('d')//a)` could be very slow if the instance is large. If so, that would probably be fatal for the async-cascade idea. But I don't know anything about the event loop etc, so hopefully the answer is no.

I tried this with a sequence of:

1. A fast function.
2. A much slower function.
3. Another fast function.

They always complete in the order queued. The same is true if there are nested calls to `requestIdleCallback`/`setTimeout`. In hindsight this makes sense: the previously requested callbacks have already been enqueued at the same priority. This is even true if the fast function in step 3 has a shorter timeout. This strongly suggests the order should actually be stable! The only way it should not be is by using timers with different priorities.

src/js/form.js
eyelidlessness: Good catch! Yes I think they should, though probably wrapped together as one function in case there are cases which shouldn't individually be async.

src/js/form.js
MartijnR: Ah, I see. I guess they are added first and then removed. I'll also play around with this in the new PR.

lognaturel (approved):

lognaturel:
It's really nice to see this isolated. Makes a lot of sense and I agree that the only real risk is something unexpected about order. With @eyelidlessness' last exploration, that seems pretty unlikely.

Let's merge this with the flags for now so it's easier to do quality assurance. With what @eyelidlessness has done I'm currently leaning towards releasing without the flag. Of course we'll know more after a QA pass. @MartijnR do you have a strong feeling one way or the other?

@jkuester this performance improvement may be another effort you're interested in. It was prompted by the fact that the relevance change we're working on does impact performance. This would maintain the same perceived performance.

MartijnR:
Thanks! Without a flag seems fine with me too. 

The forceClearNonRelevant config can be removed after #870 right?

PR #885, opened 2022-04-21
Improve test isolation

eyelidlessness:
- Perform setup/teardown in beforeEach/afterEach rather than in the describe callback
- Use stubs rather than mutating shared values like config
- Use fake timers rather than relying on timeouts to work around async delays
- Style fixes
    - Fix remaining ESLint warnings in affected files
    - Continue to move towards async/await
    - Allow Chai assertion statements

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -119,7 +119,9 @@
                 "sinon": true
             },
             "rules": {
-                "no-console": "off"
+                "no-console": "off",
+                "no-unused-expressions": "off",
+                "chai-friendly/no-unused-expressions": "warn"
             }
         }
     ]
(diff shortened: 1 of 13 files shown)


eyelidlessness:
Worth noting a couple side benefits:

- Much faster full test runs! Especially in CI.
- Much less logging noise because the stubs for those methods are actually working again.

lognaturel (approved):
I'll admit I scrubbed through this quite quickly. I'm confident there's nothing big that was removed or changed significantly. If there were a subtle error that made tests not equivalent I would probably not have caught it. Still, I think the risk is minimal.

PR #886, opened 2022-04-27
Cache local transforms

eyelidlessness:
As requested by @MartijnR in #881, breaking this out as a separate PR.

- Adds caching of transforms in `grunt transforms`
- Uses cached transforms in app.js for local forms
- Adds large WHO form for testing/validating performance improvements


app.js
```diff
@@ -27,12 +27,36 @@ if (getURLParameter('touch') === 'true') {
...
-    fetch(transformerUrl)
-        .then((response) => response.json())
-        .then((survey) => {
+    (async () => {
+        const isRemote = /^https?:\/\//.test(xform);
+        const xformURL = isRemote ? xform : `${location.origin}/${xform}`;
+        const transformerUrl = `http://${location.hostname}:8085/transform?xform=${xformURL}`;
+
+        try {
+            document.querySelector('.guidance').remove();
```
lognaturel, line 36: I can't quickly figure out what this does. Seems good that errors from it are now caught but also the error message about transform failing seems like it might be very unhelpful?

lognaturel:
I can't think of any risks here. I have spent enough time looking at it to feel like I understand the various pieces and feel confident that the caching only happens for local forms and that there's a fallback if there's no cache. I haven't fully thought through the cache busting and so a small risk remaining in my mind is the possibility of an annoying state during dev when a form source has updated but the cache hasn't for some reason.

Some small proposed changes:
- Use the new version of the VA form (https://github.com/SwissTPH/WHO-VA). If it's too different maybe it's no worth it but if it's a quick change it's good to test what's actually in use
- Add a comment in the XML form source pointing to https://github.com/SwissTPH/WHO-VA
- Does minifying the XML speed up the initial transform? By how much? If it's significant, I think we should consider checking in the minified version.

MartijnR (commented):
Nothing to add from my side. Thanks!

app.js
eyelidlessness: Sorry, most of this is more Promise -> async/await noise. Most of this logic hasn't changed, including catching errors. I have also found the error message unhelpful which is why I interpolated the error into the alert. The other substantive change of course is requesting the cached form before requesting transformer.

For what it's worth, I suspect I'm one of, if not the, most frequent users of this. 🙂

eyelidlessness:
Form is updated and comment added.

> Does minifying the XML speed up the initial transform? By how much? If it's significant, I think we should consider checking in the minified version.

It's actually a bit faster unminified!

lognaturel (approved):

lognaturel:
> It's actually a bit faster unminified!

Huh?? Is that expected?

eyelidlessness:
> > It's actually a bit faster unminified!
> 
> Huh?? Is that expected?

It's unintuitive! I'd suspect libxmljs is splitting on newlines to identify smaller chunks of work, perhaps even able to make some work concurrent?

PR #887, opened 2022-04-27
Form init performance: recalculate repeats once

eyelidlessness:
Broken out from #881 as requested by @MartijnR. Branched from #886 as that makes it easier to evaluate the performance difference.

### DIFF
--- a/src/js/form.js
+++ b/src/js/form.js
@@ -312,26 +312,34 @@ Form.prototype.init = function () {
         this.pages.init();
 
         // after radio button data-name setting (now done in XLST)
-        // Set temporary event handler to ensure calculations in newly added repeats are run for the first time
-        const tempHandlerAddRepeat = (event) => this.calc.update(event.detail);
-        const tempHandlerRemoveRepeat = () => (this.all = {});
+        let repeatsAdded = 0;
+
+        const tempHandlerAddRepeat = () => {
+            repeatsAdded += 1;
+        };
+
         this.view.html.addEventListener(
             events.AddRepeat().type,
             tempHandlerAddRepeat
         );
-        this.view.html.addEventListener(
-            events.RemoveRepeat().type,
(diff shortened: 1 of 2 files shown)


lognaturel:
A process thing just occurred to me. You could stack the PR by opening it to merge into the branch for #886, right? That way only the new commit would show up? I think it would be easier to review stacked PRs that way.

lognaturel:
Is the SOAR form only included for manual verification? Is that a thing that there's precedent for? A comment in the file about what it's there for would be helpful (commit history can get messy).

I did some looking around at existing test coverage on repeats. I was pleased to see more for nested repeats than I would have expected given @MartijnR's deep hatred thereof. 😜😜 Seems if the current repeat tests don't catch an issue with this, there's a whole form structure we haven't thought of. Certainly possible but seems unlikely to me. This is similar to what @eyelidlessness said at https://github.com/enketo/enketo-core/pull/881 (bottom).

Interested to see whether seeing the commit on its own gives @MartijnR any new ideas.

src/js/form.js
```diff
@@ -312,27 +312,35 @@ Form.prototype.init = function () {
...
+
+        // Recalculate after repeats are initialized. Previously this was performed in
+        // `tempHandlerAddRepeat`, but recalculating them all at once is significantly faster.
+        if (repeatsAdded > 0) {
+            this.calc.update({
+                allRepeats: true,
+                cloned: true,
+            });
+
+            this.all = {};
```
MartijnR, line 337: I played around with this to see if the events.RemoveRepeat() could fire without adding repeats first, and (like you) I believe that is not possible. So placing `this.all = {};` here and removing the remove handler seems 👍.

MartijnR (approved):
Thank you! This is huge, e.g. for loading records with lots of repeat instances (most common improvement, I think), forms with lots of different repeats, and forms with lots of default repeat instances.

PR #889, opened 2022-05-09
fix map with hidden geo-inputs when "hide-input" class is active

sgigh:
fix problem : It was impossible to move or select on the left of map when inputs was hidden


eyelidlessness (approved):
Once I figured out how to reproduce this, I verified it does fix the issue. Thank you!

Issue #892, opened 2022-05-18
Additions to the evaluationCascade are no longer bound to the form instance


MartijnR:
**Describe the bug**
Additions to the evaluationCascade are no longer bound to the form instance

It's a regression in https://github.com/enketo/enketo-core/commit/827c1f30a62d7e8a9093f7c9b1dac7b2b54151b4 that I should have prepared a test for, so we would have noticed. Sorry about that.

MartijnR:
A fix is to push the unchanged functions to the stack or go back to simply concatenate them as before. The challenge is to write a test.

lognaturel:
Here's the high-level gist I'm getting:
- the evaluation cascade itself is considered an integration point that clients can push arbitrary expressions into
- those expressions are expected to be evaluated in the context of the main instance
- they're not being evaluated in that context

Sort of?

lognaturel:
https://github.com/enketo/enketo-core/pull/893#issuecomment-1132075839 suggests @eyelidlessness has all the context necessary so I can  discuss with him. When reviewing the original code I remember briefly wondering what cascade additions were. It did not occur to me that they were a client-configurable thing and it feels like it's not  documented as such?

MartijnR:
Yes, it's not documented. It's a way to run a custom application without having to fork Enketo Core (just Enketo Express). It's good this broke, as I'm getting the incentive I needed to write some more of these tests (that I had promised...). :sweat_smile: I don't think documenting such advanced customizations is feasible but at least the test will provide some clarity to somebody that looks that deeply at Enketo.

MartijnR:
cloed by c69bd7d350eabf2a4de5bdbb2df0995b499292b5

PR #893, opened 2022-05-18
Fix context of evaluation cascade additions

MartijnR:
For issue #892

### DIFF
--- a/src/js/form.js
+++ b/src/js/form.js
@@ -114,11 +114,9 @@ Form.prototype = {
         const { evaluationCascadeAdditions } = this;
 
         if (evaluationCascadeAdditions.length > 0) {
-            baseEvaluationCascade.push(() => {
-                for (const fn of evaluationCascadeAdditions) {
-                    fn();
-                }
-            });
+            for (const fn of evaluationCascadeAdditions) {
+                baseEvaluationCascade.push(fn);
+            }
         }
 
         if (config.experimentalOptimizations.computeAsync) {
--- a/test/spec/form.spec.js
+++ b/test/spec/form.spec.js
@@ -1880,6 +1880,37 @@ describe('Missing external instances', () => {
     });
 });
 
+describe('Adding tasks to the default evaluation cascade', () => {
+    const defaultEvaluationCascadeAdditions =
+        Form.prototype.evaluationCascadeAdditions;
+
+    let context;
+
+    beforeEach(() => {
+        const customFn = function () {
+            context = this;
+        };
+
+        Form.prototype.evaluationCascadeAdditions = [customFn];
+    });
+
+    afterEach(() => {
+        Form.prototype.evaluationCascadeAdditions =
+            defaultEvaluationCascadeAdditions;
+    });
+
+    it('binds those tasks to the Form instance', () => {
+        const form = loadForm('thedata.xml');
+        form.init();
+        // trigger cascade
+        const b = form.view.html.querySelector('[name="/thedata/nodeB"]');
+        b.value = 'changed';
+        b.dispatchEvent(events.Change());
+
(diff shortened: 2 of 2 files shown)


test/spec/form.spec.js
```diff
@@ -1880,6 +1880,33 @@ describe('Missing external instances', () => {
...
 
+describe('Adding tasks to the default evaluation cascade', () => {
+    const defaultEvaluationCascadeAdditions =
+        Form.prototype.evaluationCascadeAdditions;
+
+    let context;
+
+    beforeEach(() => {
+        const customFn = () => {
+            context = this;
```
MartijnR, line 1891: This doesn't work because:

Screen Shot 2022-05-18 at 6 04 13 PM

If any clever ideas come to mind, I would be much obliged @eyelidlessness. Otherwise, no worries, I'll figure it out!

MartijnR: withdrawing question for now. Bunch of things wrong with that test.

src/js/form.js
```diff
@@ -114,11 +114,9 @@ Form.prototype = {
...
 
         if (evaluationCascadeAdditions.length > 0) {
-            baseEvaluationCascade.push(() => {
-                for (const fn of evaluationCascadeAdditions) {
-                    fn();
-                }
-            });
+            for (const fn of evaluationCascadeAdditions) {
+                baseEvaluationCascade.push(fn);
+            }
```
MartijnR, line 119: This is the minimal fix. I actually prefer to go back to concatenating the additions as before, but am not sure if there are reasons to use this loop instead. See alternative PR here: https://github.com/enketo/enketo-core/pull/893

test/spec/form.spec.js
```diff
@@ -1880,6 +1880,37 @@ describe('Missing external instances', () => {
...
 });
 
+describe('Adding tasks to the default evaluation cascade', () => {
+    const defaultEvaluationCascadeAdditions =
+        Form.prototype.evaluationCascadeAdditions;
+
+    let context;
+
+    beforeEach(() => {
+        const customFn = function () {
```
MartijnR, line 1890: PS. my earlier issue with `this` related to using an arrow function here.

eyelidlessness:
Thanks @MartijnR! I've taken a look, and I'm a little embarrassed that I didn't catch this! There's another regression that this addresses, where additions were no longer receiving arguments as expected.

This change (or the alternative draft PR) would certainly fix both. But I should clarify why I made the previous change to wrap the additions in a single function.

When we introduced `computeAsync`, my thinking was that we can reason about how the built in cascade methods behave if each one yields in sequence... but that we don't know how that will affect arbitrary additions. The wrapper function ensured that, when `computeAsync` is on:

1. Additions will wait until the last built-in cascade method yields, as the wrapper is also called in `callOnIdle`.

2. Once additions **are** reached, their blocking behavior is the same as it would be with `computeAsync` off.

This precaution may be unnecessary, but I think it's worth keeping until we know more about how the setting works in real world use.

I have a WIP branch where I'm integrating both the wrapping function approach with this fix restoring `this` and argument passing. Just need to finish writing a test validating the two points above still hold.

I've also reworked the test you added a bit, to use mocks rather than mutate `Form.prototype` directly. While not strictly necessary, it provides test isolation guarantees that I think are worth it.

I should have a PR open soon, but wanted to let you know I'm looking at this and my thinking on the approach.

MartijnR:
>  I'm a little embarrassed that I didn't catch this!

I didn't catch it either! And I should have written a test for this long ago....

> where additions were no longer receiving arguments as expected.

Indeed, I was about to add that the PR but will hold off.

Thanks a lot!

MartijnR:
replaced by #895

PR #895, opened 2022-05-19
Fix: ensure evaluation cascade additions receive form context and update arguments

eyelidlessness:
This is a slight variation on #893. The rationale for the difference is [explained here](https://github.com/enketo/enketo-core/pull/893#issuecomment-1132075839).

### DIFF
--- a/src/js/form.js
+++ b/src/js/form.js
@@ -114,9 +114,9 @@ Form.prototype = {
         const { evaluationCascadeAdditions } = this;
 
         if (evaluationCascadeAdditions.length > 0) {
-            baseEvaluationCascade.push(() => {
+            baseEvaluationCascade.push(function (...args) {
                 for (const fn of evaluationCascadeAdditions) {
-                    fn();
+                    fn.apply(this, args);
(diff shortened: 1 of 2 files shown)


MartijnR (approved):
That's great. Thanks!

lognaturel:
@MartijnR we have recently been leaving a few weeks between releases unless there’s something urgently needed. In general, do let us know if there’s urgency. I imagine this is a critical regression for those who depend on it.

MartijnR:
Thanks @lognaturel. This is particular one would be nice to get released in enketo-core because of the stage in OpenClinica's deployment. However, it would not be a big problem to just point to the commit in OpenClinica's fork. In that case, it would be great to just get it merged into master.

lognaturel:
@MartijnR You can still merge, right?? If you're reviewing, you should certainly feel free to take that last step. I can't do it right this moment but let's try to release today.

MartijnR:
> @MartijnR You can still merge, right??

I'm going to try! :D

Thank you!

PR #896, opened 2022-05-25
Fix: rendering unexpected geopoint widget for setvalue actions within an input

eyelidlessness:
Fixes enketo/enketo-express#431


lognaturel (approved):
We discussed that `xforms-value-changed` typically would be in the model but that wouldn't trigger this bug.

PR #900, opened 2022-06-02
Update meta selector used when setting deprecatedID

### LINKED ISSUE #890: Unexpected warning when form has nested `meta` group
**Describe the bug**

Forms in [the CHT](https://github.com/medic/cht-core) often include a nested sub-group named `meta`. I have noticed that when these forms are opened in the browser, [a warning](https://github.com/enketo/enketo-core/blob/5e1ec1bf1746664825cc4391b711596eeb205995/src/js/nodeset.js#L160) is logged:

```
Data node: /*/meta/deprecatedID with null-based index: undefined not found. Ignored.
```

**To Reproduce**

Simply init a form that has a nested group named `meta` without explicitly specifying the `deprecatedID` as a part of the model's metadata.

type | name | label::en
-- | -- | --
begin_group | text_widgets | Text widgets
begin_group |  meta | Meta Group
text | text | Text widget
end group | |  
end_group |   |  

**Expected behavior**

The expected behavior is that we do not have the warning (and the `deprecatedID` processing completes normally). It appears that the source of the problem is that when the form model is being initialized, it [creates a node for `deprecatedID`](https://github.com/enketo/enketo-core/blob/5e1ec1bf1746664825cc4391b711596eeb205995/src/js/form-model.js#L663) if one does not already exist.  However, if there are multiple `meta` groups in a form, the first element found by `this.xml.querySelector('* > meta');` might not be the top-level `meta` group for the instance (but could be a sub-group).  This causes the node for `deprecatedID` to be appended at the wrong place and then the call further down to `this.getMetaNode( 'deprecatedID' ).setVal( instanceIdExistingVal );` fails with the previously noted warning since the meta node returned by `this.getMetaNode( 'deprecatedID' )` does not actually have a `deprecatedID` child.

Now there are a couple complicating factors from the CHT that I would like to acknowledge:

When I tested a sample form out on https://getodk.org/xlsform/ I received the error `Error: There are two sections with the name meta.` (even though my `meta` group was nested and not at the top level). This leads me to think that at least at some level of the Enketo ecosystem, groups named `meta` are not supported (outside of the top level group [reserved for specifying metadata for the completed form](https://docs.getodk.org/openrosa-metadata/#completed-form-metadata)...).  So, I guess my first question is if anyone can point me to some documentation regarding this (have not been able to find anything explicit). I see quite a bit [in the spec](https://getodk.github.io/xforms-spec/#metadata) regarding the main `meta` block, but so far have not found anything that says you cannot have any other groups named `meta`.  It would be very helpful to know if the CHT is out-of-spec by allowing these groups! (Or even just if we are exposed to additional potential problems from Enketo by allowing these groups...)

Also, it is worth noting that currently the CHT does not make use of the `deprecatedID`. (We have separate logic for handling revisions to reports.) So, we are not seeing any kind of adverse behavior from this warning (at least none that we know of...). But it seemed worth raising this issue to get some additional context/thoughts here so we can understand if this could be a problem going forwards.

For what it is worth, I _think_ all that would be required to fix this processing would be to modify the selector used here: https://github.com/enketo/enketo-core/blob/5e1ec1bf1746664825cc4391b711596eeb205995/src/js/form-model.js#L663

I think just changing it to be `instance > * > meta` would prevent it from finding the nested meta groups. Let me know if this seems like a reasonable change and I am happy to raise a PR!

**Browser and OS:**

-   OS: Linux Mint 20.3
-   Browser: Chromium 101.0.4951.64
-   enketo-core: 5.18.1 (have been mainly testing with this version, but the code in question remains unchanged as of version 6.0.5)

jkuester:
Updates the selector used to pick the `meta` group element when adding an element for "deprecatedId".  Fixes an issue where the proper `meta` element is not selected if another (nested) `meta` element comes first.

Closes #890

### DIFF
--- a/test/spec/formmodel.spec.js
+++ b/test/spec/formmodel.spec.js
@@ -1721,10 +1721,11 @@ describe('merging an instance into the model', () => {
     describe('when a deprecatedID node is not present in the form format', () => {
         [
             {
+                // #890 Including second nested `meta` to validate selection logic when setting deprecatedID
                 modelStr:
-                    '<model><instance><thedata id="thedata"><nodeA/><meta><instanceID/></meta></thedata></instance></model>',
+                    '<model><instance><thedata id="thedata"><nodeA/><mygroup><meta/></mygroup><meta><instanceID/></meta></thedata></instance></model>',
                 instanceStr:
-                    '<thedata id="thedata"><meta><instanceID>7c990ed9-8aab-42ba-84f5-bf23277154ad</instanceID></meta><nodeA>2012</nodeA></thedata>',
+                    '<thedata id="thedata"><mygroup><meta/></mygroup><meta><instanceID>7c990ed9-8aab-42ba-84f5-bf23277154ad</instanceID></meta><nodeA>2012</nodeA></thedata>',
                 text: 'with meta block in default namespace, ',
             },
             // same as previous except for the namespace of the meta block and its childnodes
--- a/src/js/form-model.js
+++ b/src/js/form-model.js
@@ -660,7 +660,7 @@ FormModel.prototype.setInstanceIdAndDeprecatedId = function () {
                 'text/xml'
             ).documentElement;
             this.xml.adoptNode(deprecatedIdEl);
-            metaEl = this.xml.querySelector('* > meta');
+            metaEl = this.xml.querySelector('instance > * > meta');
             metaEl.appendChild(deprecatedIdEl);
         }
     }


test/spec/formmodel.spec.js
```diff
@@ -1721,10 +1721,11 @@ describe('merging an instance into the model', () => {
     describe('when a deprecatedID node is not present in the form format', () => {
```
jkuester, line 1721: Several of these tests fail if run with the new `modelStr` and `instanceStr` but without the new code in `form-model.js`.

MartijnR: Sorry for interrupting and to do so late but I happened to see that we already have `instanceIdEl`, so `instanceIdEl.after(deprecatedIdEl)` would be even nicer.

PR #902, opened 2022-06-05
Don't change the zoom when user click on the map

jdugh:
With a geo* question, when you click on the map, zoom change to the default zoom (if you are on zoom 19, click on map and zoom change to 15).
This PR suggest not to change the zoom ; just center the map with the new point.


eyelidlessness:
Thanks for the PR @jdugh. I took a look yesterday, and took some time looking at the surrounding code to be sure I understand the change. I'd like to try a slightly different approach, which would hopefully simplify some of the code that flows into this, and should make the zooming behavior more consistent for `setgeopoint` actions as well.

PR #905, opened 2022-06-07
Fix: preserve geopicker zoom, center map on user location on load, no error clicking pin

eyelidlessness:
Closes #902. Grouping these together, initially for visibility/feedback because they all affect the Leaflet UI and would be challenging to test and I want to get a sense of how much interest there is in adding those tests anyway.

### DIFF
--- a/src/widget/geo/geopicker.js
+++ b/src/widget/geo/geopicker.js
@@ -773,7 +773,6 @@ class Geopicker extends Widget {
         // update last requested map coordinates to be used to initialize map in mobile fullscreen view
         if (latLng) {
             this.lastLatLng = latLng;
-            this.lastZoom = zoom;
         }
 
         // update the map if it is visible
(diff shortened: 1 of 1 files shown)


src/widget/geo/geopicker.js
```diff
@@ -40,15 +40,15 @@ let searchSource =
     'https://maps.googleapis.com/maps/api/geocode/json?address={address}&sensor=true&key={api_key}';
 const googleApiKey = config.googleApiKey || config.google_api_key;
 const iconSingle = L.divIcon({
-    iconSize: 24,
+    iconSize: [16, 24],
```
eyelidlessness, line 43: This does two things:

1. Correct the width of map pins
2. Prevent an error when clicking on the pin, where Leaflet's internal state calculating the pin's offset has `NaN` for the left side.

src/widget/geo/geopicker.js
```diff
@@ -773,7 +773,6 @@ class Geopicker extends Widget {
         // update last requested map coordinates to be used to initialize map in mobile fullscreen view
         if (latLng) {
             this.lastLatLng = latLng;
-            this.lastZoom = zoom;
```
eyelidlessness, line 776: I'm not _certain_ this isn't guarded here for a reason, but it seems like the event handler more likely is the expected behavior.

src/widget/geo/geopicker.js
```diff
@@ -841,6 +840,16 @@ class Geopicker extends Widget {
                 }
             });
 
+            this.map.on('load', () => {
+                this.map.on('zoomend', (event) => {
```
eyelidlessness, line 844: Note that this is wrapped in the `load` event to ensure the initial zoom isn't treated as a user-initiated state change.

lognaturel (approved):

PR #907, opened 2022-06-15
Fix: don't reference removed repeat instance as context for relevance computations

### LINKED ISSUE #908: FomLogicError thrown on relevance on field without a control element in a repeat
A form with cascade repeats throws the following error:

`FormLogicError: Could not evaluate: not(selected( ../uccmake ,'UccMakeOther')), message: Failed to execute 'evaluate' on 'XPathEvaluator': parameter 2 is not of type 'Node'.`

The issue is caused by a mismatch in the index of a cloned repeat node. When a repeat node is added or removed the `repeatIndex` doesn't seem to sync correctly with the `repeat-number` DOM element.

eyelidlessness:
Closes #908

### DIFF
--- a/src/js/nodeset.js
+++ b/src/js/nodeset.js
@@ -246,6 +246,7 @@ Nodeset.prototype.remove = function () {
                 nodes: null,
                 repeatPath,
                 repeatIndex,
+                removed: true,
             })
         );
 
--- a/src/js/relevant.js
+++ b/src/js/relevant.js
@@ -203,8 +203,16 @@ export default {
              */
             let context = p.path;
             if (
-                getChild(node, `.or-repeat-info[data-name="${p.path}"]`) &&
-                !getChild(node, `.or-repeat[name="${p.path}"]`)
+                (getChild(node, `.or-repeat-info[data-name="${p.path}"]`) &&
+                    !getChild(node, `.or-repeat[name="${p.path}"]`)) ||
+                // Repeat instance removed for model node with no visible form control
+                (insideRepeat && repeatParent == null && options.removed) ||
+                // All repeat instances removed for model node with no visible form control, i.e. during load with `jr:count="0"`
+                (repeatPath != null &&
+                    repeatParent == null &&
+                    this.form.view.html.querySelector(
+                        `.or-repeat[name="${CSS.escape(repeatPath)}"]`
+                    ) == null)
             ) {
                 context = null;
             }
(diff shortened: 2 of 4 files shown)


eyelidlessness:
Very probably addresses #908.

src/js/nodeset.js
```diff
@@ -246,6 +246,7 @@ Nodeset.prototype.remove = function () {
                 nodes: null,
                 repeatPath,
                 repeatIndex,
+                removed: true,
```
lognaturel, line 249: Hopefully this doesn't live that long. The event is marked as for internal use so it's not a big deal to add and remove from.

In the case of a visible form control, you can look at the parent and see if it exists. In the case of a node without a visible form control, without `removed`, you may know you're in some repeat instance when trying to compute cascade stuff but you don't know whether it's been deleted.

src/js/relevant.js
```diff
@@ -203,8 +203,16 @@ export default {
...
             let context = p.path;
             if (
-                getChild(node, `.or-repeat-info[data-name="${p.path}"]`) &&
-                !getChild(node, `.or-repeat[name="${p.path}"]`)
+                (getChild(node, `.or-repeat-info[data-name="${p.path}"]`) &&
+                    !getChild(node, `.or-repeat[name="${p.path}"]`)) ||
+                // Repeat instance removed for model node with no visible form control
+                (insideRepeat && repeatParent == null && options.removed) ||
+                // All repeat instances removed for model node with no visible form control, i.e. during load with `jr:count="0"`
+                (repeatPath != null &&
```
lognaturel, line 211: Is `repeatPath != null` the same as `insideRepeat`?

src/js/relevant.js
```diff
@@ -203,8 +203,16 @@ export default {
...
+                    repeatParent == null &&
```
lognaturel, line 212: `repeatParent == null` is only true if the node had no visible form control

src/js/relevant.js
lognaturel: `insideRepeat` should be used here (it is equivalent to `repeatPath != null && p.path.startsWith(`${repeatPath}`);`).

lognaturel:
Closing in favor of minor changes at #909. Opening another PR instead of pushing to this one to preserve history.

Since @eyelidlessness is out, I'll let #909 sit for ~a day to see if there are any comments and then will merge and release so we can fix the regression.

PR #909, opened 2022-06-20
Fix: don't use removed repeat instance as context for relevance

### LINKED ISSUE #908: FomLogicError thrown on relevance on field without a control element in a repeat

lognaturel:
Closes #908, replaces #907

This uses dedicated forms to narrowly reproduce and test the problem. @eyelidlessness may also introduce a smoketest on the real form separately.

It also slightly simplifies the test according to https://github.com/enketo/enketo-core/pull/907#discussion_r899398203

### DIFF
--- a/src/js/nodeset.js
+++ b/src/js/nodeset.js
@@ -246,6 +246,7 @@ Nodeset.prototype.remove = function () {
                 nodes: null,
                 repeatPath,
                 repeatIndex,
+                removed: true, // Introduced to handle relevance on model nodes with no form controls (calculates)
             })
         );
 
--- a/src/js/relevant.js
+++ b/src/js/relevant.js
@@ -203,8 +203,16 @@ export default {
              */
             let context = p.path;
             if (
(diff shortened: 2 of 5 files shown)


lognaturel:
As written in #907, I'll leave this up for ~a day to see if there are any other comments but since it doesn't make substantive changes to #907 I'll consider this my review and then get the patch out.

yanokwa (approved):
I didn't look at the tests. I confirmed that nodeset.js was not changed and that the refactoring of relevant.js LGTM

PR #911, opened 2022-06-24
Prepare v6.1.2 release

lognaturel:
I did not update dependencies. I think we should update the release instructions to say we usually don't for point releases. It seems too risky to me -- we won't do any QA on patches typically.

I did take a quick look to see whether anything was high severity. https://github.com/chalk/ansi-regex/pull/37 is marked as high severity but it seems... not even kind of?


yanokwa (approved):
LGTM

PR #914, opened 2022-08-01
fix toggle or-appearance-compact with dynamic nesteed repeat group

jdugh:
If you have a nested repeat group, `this.view.$.find('.or-group > h4').on('click', function () [...] ` is not triggered on the dynamic add group.
In this example :
image

At first I have 1 parent repeat group with 1 child repeat group.
I add a parent repeat group.
I can't click on the h4 of the second child repeat group to compact it

### DIFF
--- a/src/js/form.js
+++ b/src/js/form.js
@@ -1050,7 +1050,7 @@ Form.prototype.setEventHandlers = function () {
         this.output.update();
     });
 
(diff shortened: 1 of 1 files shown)


eyelidlessness (approved):
Nice, simple fix. Thanks!

PR #916, opened 2022-09-26
Fix timepicker in Finnish locales

eyelidlessness:
Fixes enketo/enketo-express#453. Opening as draft for now until the following are complete (unless we determine not to do the remaining work):

- [x] Use the [`Intl.DateTimeFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) API rather than regex to determine time structure.
- [x] Use a locale-specific date string to determine whether a time *string* has a meridian component. This is necessary because you can't `new Date('01:00 pm')`.
- [x] Detect locale changes and updates the locale state used for the above accordingly.
- [x] Detect locale change to update widget state.
- ~~[ ] Format widget display state with locale-specific punctuation rather than hard coding the separator as `:`~~

### DIFF
--- a/test/spec/format.spec.js
+++ b/test/spec/format.spec.js
@@ -1,62 +1,76 @@
-import { format, time } from '../../src/js/format';
+import { time } from '../../src/js/format';
 
 describe('Format', () => {
     /** @type {import('sinon').SinonSandbox} */
     let sandbox;
 
-    /** @type {string} */
-    let localeOverride;
+    [
+        {
+            locale: 'en-US',
+            hour12: true,
+            am: 'AM',
+            pm: 'PM',
+        },
 
-    beforeEach(() => {
-        sandbox = sinon.createSandbox();
+        /**
+         * Note (2022-01-31) — This fixture previously used the `zh` locale.
+         * Tests for this fixture pass in Chrome 97 (current) and Firefox 95
+         * (current - 1), but fail in Firefox 96 producing English-language
+         * 24-hour time. The expected behavior is produced for `zh-hk`.
+         */
+        {
+            locale: 'zh-HK',
+            hour12: true,
+            am: '上午',
+            pm: '下午',
+        },
 
-        sandbox.stub(format, 'locale').get(() => localeOverride);
-    });
+        {
+            locale: 'ar-EG',
+            hour12: true,
+            am: 'ص',
+            pm: 'م',
+        },
+        {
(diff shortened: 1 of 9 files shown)


lognaturel:
Nice! Do you see a big advantage to doing the last two points in this PR vs treating them as separate issues?

Both seem nice to have but not critical to me so it all depends on whether you think they’re quick or not.

eyelidlessness:
It was pretty straightforward to handle the language change. I decided to hold off on the separator though because it occurs to me there's probably other issues with that as well. Anyway, I think this is now ready for review!

test/spec/format.spec.js
```diff
@@ -1,62 +1,76 @@
-import { format, time } from '../../src/js/format';
+import { time } from '../../src/js/format';
 
 describe('Format', () => {
     /** @type {import('sinon').SinonSandbox} */
     let sandbox;
 
-    /** @type {string} */
-    let localeOverride;
+    [
```
lognaturel, line 7: I'm surprised not to see any additions here! Is there a test that would have failed prior to the change?

lognaturel (commented):
Looks good to me! I'm just having trouble identifying a test that would have failed previously. Worth doing a sanity check with Express before merging?

test/spec/types.spec.js
```diff
@@ -7,6 +8,17 @@ import types from '../../src/js/types';
 describe('Types', () => {
     describe('time', () => {
         describe('meridian convertor', () => {
+            /** @type {Function} */
+            let teeardownTimeLocalization;
```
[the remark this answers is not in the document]
eyelidlessness: Nope, typo + copypasta 🙃

test/spec/format.spec.js
eyelidlessness: Good call, I meant to actually add the case from the issue! I was focused on getting the setup/te~~e~~ardown switched over and forgot.

lognaturel (approved):
Looks great. @eyelidlessness mentioned some risk around date/times coming in from new submissiosn vs. existing submissions vs. defaults and whether those are all localized at the same time. As far as I can tell, the interaction with all of those is unchanged from the original regex-based solution so it doesn't seem like there's a lot of risk there to me. Still, I've made a note for QA to do some additional verification once it's available in Express!

PR #918, opened 2022-09-28
update choice-name to also support select-multiple

jdugh:
Actually, choice-name function support only select-one (or select-multiple if only one element is selected).
I suggest to change the code for split the `name` with a space (the list separator), and loop over it for find all labels.
At the end, join it with a space.
image

What do you think about this ?

### DIFF
--- a/src/js/form.js
+++ b/src/js/form.js
@@ -519,63 +519,72 @@ Form.prototype.replaceChoiceNameFn = function (
         const params = choiceName[1];
 
         if (params.length === 2) {
-            let label = '';
-            const value = this.model.evaluate(
+            const labelArray = [];
+            const values = this.model.evaluate(
                 params[0],
                 resTypeStr,
                 context,
                 index,
                 tryNative
             );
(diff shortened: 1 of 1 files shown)


lognaturel:
Thanks for sharing this idea. Please see [this ODK forum thread](https://forum.getodk.org/t/summarizing-previous-answers-from-a-multiple-select-question-in-a-note/21721/8) and see if ideas there could help you. If not, it would be helpful to get a description of your underlying need. Is it like the original user who would like to ask for confirmation? Something else?

Ideally we would come up with a solution that would give the form designer control over the separator. We're very close with the solution I shared in that thread but need a way to address translations.

jdugh:
I had looked for a solution but I had not found this one (the join).
That's exactly what I want, with the extra configurable separator!

lognaturel:
Oh great, glad to hear it, @jdugh! Consider posting these kinds of questions or ideas on the ODK forum so they get more eyes. 👀

PR #926, opened 2022-11-24
Fix: validate number input values

### LINKED ISSUE #792: Decimal fields accept any combination of , . - as inputs but submit empty

eyelidlessness:
Fixes #792, closes #810.

The underlying problem is that the native `number` input automatically blanks invalid `value`s, and Enketo was not considering native form validation cases. Chrome prevents users from entering invalid values in the first place, but Firefox and Safari do not. So while a number input may display an invalid value, the underlying view and model data is blank.

This is a narrow fix, only addressing native validation of `number` inputs which are not handled by other widgets. It may be a good idea to extend native validation logic to forms generally, but I'm not sure how much risk that would entail.

This change also moves number input handling into a widget. This isn't strictly necessary, but it's a direction I'd like to move for non-widget logic where a widget would make sense.

Lastly, a quick note about a couple changes in logic from the current `mask` implementation: rather than attempting to detect copy/cut/paste keyboard commands, this now allows _any_ cmd- (meta-) or ctrl- keyboard shortcut. And the paste event would be redundant, because validation is handled in an `input` event which is also fired on cut and paste.


lognaturel:
I really don't understand what's happening with the build:

```
Run xvfb-run --auto-servernum npm run test-browsers

> enketo-core@6.1.3 test-browsers /home/runner/work/enketo-core/enketo-core
> cross-env NODE_OPTIONS='--max-old-space-size=8192' grunt karma:browsers

Running "karma:browsers" (karma) task
28 11 2022 13:19:11.625:ERROR [launcher]: Cannot start Firefox
	Can not find the binary firefox
	Please set env variable FIREFOX_BIN
```

Things work locally. I don't see anything in this branch which would change the environment related to Firefox. I pushed a new branch off `master` to see whether it would fail and it doesn't (https://github.com/lognaturel/enketo-core/actions/runs/3568190243/jobs/5996779996).

I'm going to merge. It seems like there might be something about this CI run that's not right and that's not reset by a full re-run.

lognaturel (approved):

lognaturel:
> I'm going to merge. It seems like there might be something about this CI run that's not right and that's not reset by a full re-run.

Sigh, I was wrong and it fails on `master` as well.

lognaturel:
Same commit works on `master` on my branch: https://github.com/lognaturel/enketo-core/commit/fd21204caab56250c70c08de698f82a693480179

The one possibly substantive difference I notice is that `npm ci` runs on my branch. I'm going to try to do the release prep which bumps dependencies and see if that gets things working here.

PR #927, opened 2022-11-28
Prepare 6.1.4 release

lognaturel:
I only updated vulnerable packages to minimize risk.


lognaturel:
Replaced by #928 which will explicitly install Firefox.

Issue #924, opened 2022-11-02
Signature and draw widget are considered to have a value after it has been emptied/reset


MartijnR:
**Describe the bug**
When adding a signature and then removing it, it is considered to have a value after 2 successive validations

**To Reproduce**
Steps to reproduce the behavior. 

1. Open [sig_widget.xml.txt](https://github.com/enketo/enketo-core/files/9923121/sig_widget.xml.txt) in enketo-core
2. Add a signature
3. Reset the signature widget
4. Click Validate and confirm correct 'required' validation error
5. Click Validate again and see that the 'required' validation error is no longer there

**Expected behavior**
For an emptied signature widget to be considered empty until a new signature is drawn.

**Additional context**
I'll do some further investigation and will edit/update this issue.

(P.S. the delay in signature widget populating is also a little odd because of validation errors appearing and disappearing with a delay but haven't heard complaints about this yet.)

lognaturel:
Maybe some relationship with https://github.com/enketo/enketo-express/issues/339?

MartijnR:
Notes (to self mostly):

### cause: canvas blur event handler firing after validation

* The canvas automatically gets focus if it's the first invalid question on the form. 
* The widget is not able to make a distinction between an empty canvas and an edited canvas and presumably returns the empty image as dataUrl. 
* You can reproduce the bug without validation too by using the keyboard to toggle focus/blur on a previously-cleared canvas.

### fixes:

1. Adding a (or using an already-added) mechanism to make the widget aware of whether the user added any strokes and if not, return an empty value (which is an empty string I believe).

lognaturel:
Closed by #931

PR #931, opened 2022-11-30
Fix/drawing widget blur 924

MartijnR:
closes issue #924 

I wasn't able to create a test for this. I'm guessing this is because canvas doesn't work in a headless browser. Fwiw, I tried this to get to the first step (just before resetting):

```js
describe('draw widget', () => {
    /** @type {import('sinon').SinonSandbox} */
    let sandbox;

    /** @type {SinonFakeTimers} */
    let timers;

    beforeEach(() => {
        sandbox = sinon.createSandbox();
        timers = sandbox.useFakeTimers();
    });

    afterEach(() => {
        timers.runAll();
        timers.clearTimeout();
        timers.clearInterval();
        timers.restore();
        sandbox.restore();
    });

    it(`does not create an image of an empty canvas when the canvas looses focus (blurs)
     after it has been reset at least once`, () => {
        const form = FORM2;
        const fragment = document.createRange().createContextualFragment(form);
        const control = fragment.querySelector(DrawWidget.selector);
        new DrawWidget(control);
        const canvas = fragment.querySelector('canvas');

        const touch1 = new Touch({
            identifier: 123,
            target: canvas,
        });

        const touchEvent1 = new TouchEvent('touchstart', {
            touches: [touch1],
        });

        const touch2 = new Touch({
            identifier: 234,
            target: canvas,
        });

        const touchEvent2 = new TouchEvent('touchend', {
            touches: [touch2],
        });

        canvas.dispatchEvent(touchEvent1);
        canvas.dispatchEvent(touchEvent2);
        canvas.dispatchEvent(new Event('blur'));

        expect(control.value).to.equal('signature.png');
    });
});
```

The fix is straightforward though, so maybe this is acceptable without a test? :shipit: 

I also considered, as an alternative fix, to get rid of the blur handler which would mean removing or much reducing the on-stroke-end delay in saving an image. However, it was considered out of scope to investigate whether this would cause (performance) issues on low-end devices.

### DIFF
--- a/src/widget/draw/draw-widget.js
+++ b/src/widget/draw/draw-widget.js
@@ -135,15 +135,19 @@ class DrawWidget extends Widget {
             this._handleFiles(existingFilename);
         }
 
-        // We built a delay in saving on stroke "end", to avoid excessive updating
-        // This event does not fire on touchscreens for which we use the .hide-canvas-btn click
-        // to do the same thing.
+        // This listener (I think) serves to capture a drawing when the submit button is clicked within [DELAY]
+        // milliseconds after the last stroke ended. Note that this could be the entire drawing/signature.
         canvas.addEventListener('blur', () => {
             // If the canvas is empty, calling blur would result in an empty image file being created.
+            // Validation logic and traversing the form by keyboard could trigger those blur events on empty canvases.
             if (this.pad && !this.pad.isEmpty()) {
                 this._forceUpdate();
             }
         });
+
+        // We built a delay in saving on stroke "end", to avoid excessive updating
+        // This event does not fire on touchscreens for which we use the .hide-canvas-btn click
+        // to do the same thing.
 
         this.initialize = fileManager.init().then(() => {
             that.pad = new SignaturePad(canvas, {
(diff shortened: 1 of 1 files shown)


MartijnR:
very dirty commits, so needs squash imo

eyelidlessness (commented):
I've verified that this works as intended/described, and I believe the change would address any remaining issues with enketo/enketo-express#339.

That said, I can't help but wonder whether the delay is necessary. I imagine it's is intended to throttle updates for rapid input e.g. a non-contiguous signature?

Currently, standard form inputs don't update their value until a `change` event (on `blur` in most cases, but sometimes more frequently e.g. for number input steps). There's certainly value in updating the UI to set the appropriate state for the widget's reset and download buttons, as well as the filename value which could trigger related computations. But could the update to the binary state be deferred until `blur`[^1]?

This would be consistent with the mobile/touch behavior, probably even allowing us to eliminate the distinction (except perhaps the modal presentation if that's still preferred). I _think_ that eliminating the delay but deferring the binary state change until `blur`[^1] would have the same (or better) throttling impact, without any delay, thereby preventing race conditions (which are at the root of these issues).

[^1]: Or any other form state finalizing event, like `submit` or `reset` or validation. These are generally prevented while a field is still focused, but I'm not sure we should assume that will or should always be the case.

lognaturel (approved):
Let's get this out now since it fixes a real issue!

@eyelidlessness will file an issue for future follow-up.

PR #932, opened 2022-12-05
Fix: allow arbitrary precision decimal values

eyelidlessness:
For future reference: this relies on explicit behavior in enketo-transformer which sets `step="any"` on decimal inputs. The affected tests have been updated accordingly to use the transformed result, rather than arbitrary HTML string literals.

### DIFF
--- a/test/forms/number-input-widgets.xml
+++ b/test/forms/number-input-widgets.xml
@@ -1,0 +1,30 @@
+<?xml version="1.0"?>
+<h:html xmlns="http://www.w3.org/2002/xforms"
+    xmlns:ev="http://www.w3.org/2001/xml-events"
+    xmlns:h="http://www.w3.org/1999/xhtml"
+    xmlns:jr="http://openrosa.org/javarosa"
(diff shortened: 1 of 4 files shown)


test/forms/number-input-widgets.xml
```diff
@@ -0,0 +1,30 @@
...
+            </instance>
+            <bind nodeset="/data/int" type="int"/>
+            <bind nodeset="/data/decimal" type="decimal"/>
+        </model>
+    </h:head>
+    <h:body>
+        <input ref="/data/int">
+            <label>Integer</label>
+        </input>
+        <input ref="/data/decimal" step="0.1">
```
lognaturel, line 26: Step is not part of the form spec and maybe best not to have it here for consistency?

eyelidlessness: Oops, good catch! I put that in to see what Transformer's behavior would be, forgot to remove it.

lognaturel (approved):

PR #935, opened 2022-12-15
Fix calculation of nested repeat counts in >1 outer repeat instances

### LINKED ISSUE #934: Nested repeats with repeat_count are broken
**Describe the bug**
A nested repeat group using `repeat_count` does not repeat the inner group's questions after the first outer instance.

**To Reproduce**

**survey**

| type         | name         | label        | repeat_count | calculation  |
| ---          | ---          | ---          | ---          | ---          |
| begin_repeat | outer_repeat | Outer Repeat | 2            |              |
| calculate    | p1           |              |              | position(..) |
| text         | q1           | Q${p1}       |              |              |
| begin_repeat | inner_repeat | Inner Repeat | 2            |              |
| calculate    | p2           |              |              | position(..) |
| integer      | q2           | Q${p1}.${p2} |              |              |
| end_repeat   |              |              |              |              |
| end_repeat   |              |              |              |              |

[nested-repeat-count.xlsx](https://github.com/enketo/enketo-express/files/10200303/nested-repeat-count.xlsx)

**Expected behavior**

The inner group questions are repeated in all outer group instances.

**Screenshots**
 
Expected behavior observed in EE `3.0.5`

Broken behavior in EE `5.0.0`:

- OS: macOS
- Browser: Firefox

**Additional context**
Issue present since at least EE `4.1.1`.

eyelidlessness:
Fixes #934

### DIFF
--- a/src/js/form.js
+++ b/src/js/form.js
@@ -347,46 +347,96 @@ Form.prototype.init = function () {
         // before repeats.init so that template contains role="page" when applicable
         this.pages.init();
 
-        // after radio button data-name setting (now done in XLST)
-        let repeatsAdded = 0;
+        // Builds a cache of known repeat path prefixes `repeat.init`.
+        // The cache is sorted by length, longest to shortest, to ensure
+        // that lookups using this cache find the deepest nested repeat
+        // for a given path.
+        this.repeatPathPrefixes = Array.from(
+            this.view.html.querySelectorAll('.or-repeat-info')
+        )
+            .map((element) => `${element.dataset.name}/`)
+            .sort((a, b) => b.length - a.length);
 
-        const tempHandlerAddRepeat = () => {
-            repeatsAdded += 1;
-        };
+        if (this.repeatPathPrefixes.length > 0) {
+            const nestedRepeats = Array.from(
+                this.view.html.querySelectorAll('.or-repeat .or-repeat')
+            );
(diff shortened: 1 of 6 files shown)


src/js/form.js
```diff
@@ -347,46 +347,96 @@ Form.prototype.init = function () {
...
+                    this.all = {};
+                    didRecalculate = false;
+                }
+            };
 
-        this.view.html.removeEventListener(
-            events.AddRepeat().type,
-            tempHandlerAddRepeat
-        );
+            if (recalculationPaths.length > 0) {
```
eyelidlessness, line 406: This if should almost certainly be removed

src/js/form.js
```diff
@@ -347,46 +347,96 @@ Form.prototype.init = function () {
...
+                    addRepeatType,
+                    tempHandlerAddRepeat
+                );
+                this.view.html.removeEventListener(
+                    removeRepeatType,
+                    tempHandlerRemoveRepeat
+                );
+            }
+
+            if (!didRecalculate) {
```
lognaturel, line 431: Remove. Might lead to extra work being done but there's some risk in having it there.

lognaturel (approved):
@eyelidlessness and I did an interactive review and I can follow code moves and newly-introduced conditions. I see that it does turn off the optimization from #887 for nested repeats and their parents.

PR #941, opened 2022-12-21
Fix: don't clear decimals when switching French -> English

eyelidlessness:
Moving `export default` below should also fix an oddity causing [this docs page](https://enketo.github.io/enketo-core/module.exports.html) to be generated.


lognaturel (approved):

lognaturel:
Did another quick round of manual verification and can't think of any other cases!

PR #946, opened 2023-01-24
Removed IE11 remnants

### LINKED ISSUE #945: remove IE11 remnants
**Describe the bug**
OpenClinica has dropped support. I think it may be okay to clean up the IE11 remnants from the code now.

MartijnR:
Closes #945 

I relied on the tests to reveal regressions.


MartijnR:
Some browser test failed. I presume in the 2 year old MS Edge. I updated all browsers to the latest stable version which resolved it. I didn't do any further investigation. If you'd like me to find out exactly which browser and version failed, let me know though!

MartijnR:
Just realized the esbuild config update made no sense as a fix because it modernized the syntax of the JS bundle. Not sure why that worked….

src/js/form-model.js
```diff
@@ -584,20 +505,8 @@ FormModel.prototype.mergeXml = function (recordStr) {
 
     // Remove the primary instance childnode from the original model
     this.xml.querySelector('instance').removeChild(modelInstanceChildEl);
```
eyelidlessness, line 507: No need to change now, noting for posterity: I think by using `adoptNode`, this can be removed.

eyelidlessness (approved):

PR #952, opened 2023-02-15
removed: analog-scale widget, closes #950

### LINKED ISSUE #950: Remove Analog Scale widget?
The analog scale widget was developed for a particular clinical trial use case (for OpenClinica). It has to rigidly look a particular way in a particular form to be certified for certain trials. I think my rationale for putting it in Enketo Core (long ago) was that others could perhaps benefit from it.

However, likely nobody outside OpenClinica knows about it. It was never added to the default widgets configuration in Enketo Express either (and not to ODK Collect).

It might be more efficient for OpenClinica to add/move this widget to their fork of Enketo Express and continue working on it there. 

Would anybody have any objections to removing this widget from Enketo Core (to avoid confusion)? Or is anybody excited about this potential code cleanup?

MartijnR:
There are a few places where a query selector or a css selector includes analog-scale. It would be great to keep those for now unless you have an approach in mind that would allow those cases to be addressed (in the fork).


src/widget/number-input/number-input.js
```diff
@@ -34,6 +34,7 @@ class NumberInput extends Widget {
 
         const question = input.closest('.question');
 
+        // analog-scale is included as a courtesy to OpenClinica
```
eyelidlessness, line 37: This (and others) seems fine for now, but I'd like to think about what an extensible interface for these cases might look like. Partly because these will become harder to track and account for with the OpenClinica code removed (i.e. I only knew to add this here by looking at other usage of number inputs). But mainly because CSS selectors and similar DOM state lookups are _by far_ our biggest performance challenge in Core.

eyelidlessness (approved):
Always nice to see some code removed! I left one note for posterity about future extensibility. I think it might also be good to think about ways these could be conditional at build time, but that would likely also involve changes in Express.

src/widget/number-input/number-input.js
MartijnR: > but I'd like to think about what an extensible interface for these cases might look like

Yes, would definitely be best. 

Thanks!

Issue #956, opened 2023-03-01
Print view displays inline span elements inside a label as blocks if the question is a select or select_one


MartijnR:
**Describe the bug**
See the screenshots below of a simple form that contains 2 questions with identical labels. The print view displays the labels differently if the question type is a select or select_one.

**To Reproduce**

1. add [inline-span-labels-print.xml.txt](https://github.com/enketo/enketo-core/files/10864406/inline-span-labels-print.xml.txt) to  `test/temp`
2. open form: http://localhost:8005/?xform=inline-span-labels-print.xml.txt
3. Observe correct identical labels in screen view

<img width="648" alt="Screenshot 2023-03-01 at 1 51 52 PM" src="https://user-images.githubusercontent.com/627350/222236746-676823b2-5a8c-4041-888c-d2fd0be703b4.png">

4. Open print view with CMD-P  [^1] and observe how inline spans are displayed on new lines in the 2nd question

<img width="741" alt="Screenshot 2023-03-01 at 1 53 04 PM" src="https://user-images.githubusercontent.com/627350/222236993-a585fcbf-8b20-47e4-8b5f-fa5b62bf2470.png">

[^1]: Alternatively change print style attribute `media="print"` to `media="all"` in index.html **and restart app** to load in the print view right away.

**Expected behavior**

For the print view to display the label as the screen view (in step 3).

PR #957, opened 2023-03-01
fixed: undesired newlines in print view in select/select1 labels, clo…

MartijnR:
…ses #956

As-minimal-as-possible change to avoid any display issues. I think the risk of regressions is very small (and only in print views).

I also checked if this worked in:

- the print view of the Formhub theme
- the print view of the Grid theme
- the interaction with a hint in both themes

Screenshot 2023-03-01 at 2 01 40 PM

### DIFF
--- a/src/sass/core/_print.scss
+++ b/src/sass/core/_print.scss
@@ -147,7 +147,7 @@ input[type='checkbox'] {
             padding: 0;
         }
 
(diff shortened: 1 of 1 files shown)


eyelidlessness (approved):
I almost approved this from my phone because it's so obviously correct, but I did verify it works 🙂

PR #958, opened 2023-03-16
Remove autogenerated docs

lognaturel:
It's not clear what the intended audience of these docs was. Most users should focus on the README/tutorials to learn about usage and extension points. Most of what was included in the jsdoc is not intended to be a public API to develop against.

If someone really needs this, they can let us know why.


.eslintrc.json
```diff
@@ -9,18 +9,11 @@
...
         "sourceType": "module",
         "ecmaVersion": 2020
     },
-    "settings": {
-        "jsdoc": {
-            "tagNamePreference": {
-                "returns": "return"
-            }
-        }
-    },
```
eyelidlessness, line 23: I'm *perfectly fine* with removing the `settings` portion, and if it turns out we're not depending on the plugin removing that too. But to be clear, this is not part of the doc generation.

tutorials/README.md
```diff
@@ -0,0 +1,7 @@
+# Enketo Core tutorials
+
+* [Getting started](getting-started.md)
+* [Configuration](configuration.md)
+* [Development](development.md)
+* [Events](events.md)
+* [Creating/extending widgets](widgets.md)
```
eyelidlessness, line 7: Why not just let GitHub provide the navigation?

.eslintrc.json
lognaturel: Yeah, I guess we might as well keep that consistent. Will undo.

tutorials/README.md
lognaturel: I... don't know?

eyelidlessness (approved):

PR #960, opened 2023-03-17
Fix: preserve high fidelity image data when resizing draw-widget canvas

eyelidlessness:
Fixes enketo/enketo-express#447.

### Primary bug fix

Regardless of the size of the viewport/`canvas` element:

- Drawing data will scale up and down on resize without losing fidelity
- Base images for annotations will scale up and down to their intrinsic size
- The raster data which will ultimately be included as submission attachments will scale, with full drawing fidelity, to:
  - For annotations, the intrinsic size of the base image.
  - For non-annotations, the maximum display size of the `canvas` element (regardless of the size it was drawn). This will be theme-specific, determined by `max-width` styles specified on either the `canvas` element itself or its nearest descendant specifying it.

To accomplish this, this change introduces a new `ResizableSignaturePad` subclass of the class provided by the underlying `signature_pad` library. That subclass now [re]implements these behaviors previously in `draw-widget.js`:

- Resizing and redrawing to the `canvas` element when resized in the view
- Assigning base images for annotations
- Compositing annotation drawing data with annotated base images
- Dispatching events to update `DrawWidget`'s state (thereby updating the form's model state as well)
- Producing an opaque object URL to reference the current bitmap state (e.g. for downloads)

Those behaviors being isolated, several parts of `DrawWidget` have become simplified, deferring to `ResizableSignaturePad`.

### Related changes

Besides preserving image fidelity, this change introduces some performance improvements, which in turn allows improved ergonomics for integration in e.g. Enketo Express. There's some additional discussion of this in code comments. Briefly, these improvements come from using these standard web APIs:

- `IntersectionObserver` to determine visibility of each individual `canvas` element, and defer certain work when not visible
- `ResizeObserver` to detect when each individual `canvas` element needs to be scaled and redrawn
- `ImageBitmap` to allow synchronous redraws of annotation base images
- `URL.revokeObjectURL` called with stale object URLs as new ones are created, to free memory as image state is changed

Aside from those APIs, quite a lot less work is being performed on resize/visibility change:

- Scaling drawing data, for both the view and the rasterized image, operates only on the in-memory data structures used by the underlying library
- The `canvas` element is redrawn, but no state is changed until the user either changes the base image or the drawing state

All of these improvements together are significant enough that throttling in `DrawWidget` is no longer necessary (i.e. state will update synchronously, and downstream users such as Enketo Express will no longer need to accommodate this particular delay). And since `ImageBitmap` allows synchronous redraws, resizes perform perceptibly faster even without any throttling there.


lognaturel (approved):
First, I think the outcome here is great: 
* in the case of annotating an existing image, it's natural to expect that the submitted image would have the same size as the input image
* in the case of drawing or signing a blank canvas, it's natural to expect that all submissions would have the same fixed size

I spent some time looking at the tests and the general structure. The tests look thorough. It's a little too bad that the end-to-end tests don't work across browsers but I don't think it's that important. Pulling functionality out of the widget with an eye to making a possibly more general thing is consistent with our broader goals.

I have not very carefully reviewed `ResizableSignaturePad`. There's some risk of small unintended changes in behavior but given the verification @eyelidlessness has done and the level of testing, I think those are low.

I'd like to merge this immediately, release, and get it updated in Express so that both this and the Transformer change can be exercised on XLSForm online.

eyelidlessness:
Thanks for the quick review @lognaturel! Per our quick Slack discussion, I'm going to do a quick check of integration with Express, to validate that binary defaults/instance submissions work as expected. If so, I'll go ahead and merge.

PR #969, opened 2023-04-21
Fix: focus first active input/widget on load, page navigation, user-added repeat

eyelidlessness:
Fixes enketo/enketo-express#277. The tests aren't exhaustive e.g no tests for non-user-added repeats, how relevance impacts determination of whether a field is active. I'd be happy to add any test scenarios but wanted to get this into review first. It's also probably not an exhaustive fix. I hadn't originally caught the select toggle button case, nor the readonly cases (which is why they're used so much in the tests), and I would be surprised if there aren't other edge cases I haven't encountered.

### DIFF
--- a/src/js/dom-utils.js
+++ b/src/js/dom-utils.js
@@ -463,6 +463,34 @@ class MutationsTracker {
     }
 }
 
+/** @type {HTMLElement | null} */
+let scrollIntoViewTarget = null;
+
+const intersectionObserver = new IntersectionObserver((records) => {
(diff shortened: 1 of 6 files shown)


lognaturel:
It's a little bit surprising to me that the first question in the new repeat jumps to the top of the screen but I do think it's a big improvement for keyboard navigation. Is the jump intentional? Is it straightforward to jump/not jump or is that currently a part of what "focus" means?

lognaturel (approved):

PR #971, opened 2023-05-03
Performance improvements: NumberInput widgets, `excludeNonRelevant`

eyelidlessness:
This change includes performance optimizations to perform fewer redundant operations in:

- `NumberInput` widgets (`DecimalInput`, `IntegerInput`). Their original implementation introduced a performance regression for forms with a large quantity of numeric inputs. Some of the pertinent change may benefit other widgets, i.e. they may be able to take advantage of `globalInit`, which is currently being used to cache the global form element rather than repeatedly querying for it, and to cache the localization behaviors in `DecimalInput`. Other changes move more decimal-specific logic from `NumberInput` to `DecimalInput` so that logic isn't performed needlessly by `IntegerInput`.

- Relevance changes when using the `excludeNonRelevant` setting. Prior to this change, the implementation would perform some unnecessary DOM operations (rewriting blank values to nodes which were already blank), and often many unnecessary downstream computations (dispatching `DataUpdate` events for those same nodes). This change will particularly benefit forms where a relevance change cascades to a large number of blank/unfilled elements.

### DIFF
--- a/src/widget/number-input/number-input.js
+++ b/src/widget/number-input/number-input.js
@@ -1,7 +1,25 @@
 import { t } from '../../js/fake-translator';
-import events from '../../js/event';
-import inputModule from '../../js/input';
 import Widget from '../../js/widget';
+
+/** @type {WeakMap<HTMLInputElement, HTMLElement>} */
+const questionsByInput = new WeakMap();
+
+/**
+ * @param {HTMLInputElement} input
+ */
+const getQuestion = (input) => {
+    let question = questionsByInput.get(input);
+
+    if (question == null) {
+        question = input.closest('.question');
+
+        if (question != null) {
+            questionsByInput.set(input, question);
+        }
+    }
+
+    return question;
+};
 
 /**
  * @abstract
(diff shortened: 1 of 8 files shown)


lognaturel:
Changing the form language to Arabic changes the numerals for decimal but not for integer, now.

src/widget/number-input/number-input.js
```diff
@@ -1,93 +1,173 @@
...
 import events from '../../js/event';
-import inputModule from '../../js/input';
+import { t } from '../../js/fake-translator';
 import Widget from '../../js/widget';
 
+/** @type {Set<NumberInput>} */
+let numberInputInstances = new Set();
+
+/** @type {WeakMap<HTMLInputElement, HTMLElement>} */
+let questionsByInput = new WeakMap();
```
lognaturel, line 9: Having this map here feels weird to me but @eyelidlessness has confirmed that the query to get the input container (`.question`) is quite expensive and ends up being run all the time. This especially affects forms with number inputs in repeats.

lognaturel (approved):

PR #973, opened 2023-05-05
Fix: localized numeral entry by keyboard in number input widgets

### LINKED ISSUE #972: Digits that are not 0-9 (e.g ١ -٩ in arabic) can't be entered in number inputs
As of at least the change to use widgets for numeric input and perhaps before, direct input of localized digits that are not 0-9 is not allowed. Changing the form language does show the correctly localized digits and localized digits can be pasted in.

eyelidlessness:
Fixes #972.

### DIFF
--- a/src/widget/number-input/decimal-input.js
+++ b/src/widget/number-input/decimal-input.js
@@ -27,28 +27,28 @@ const getDecimalCharacters = (languages) => {
     return characters;
 };
 
-/** @type {Map<string, RegExp>} */
-let characterPatternsByLanguage = new Map();
+/** @type {Map<string, Set<string>>} */
+let validCharactersByLanguage = new Map();
 
 /**
  * @param {string[]} languages
  */
-const getCharacterPattern = (languages) => {
+const getValidCharacters = (languages) => {
     const [language] = languages;
-    let characterPattern = characterPatternsByLanguage.get(language);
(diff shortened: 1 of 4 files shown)


src/widget/number-input/decimal-input.js
```diff
@@ -27,28 +27,28 @@ const getDecimalCharacters = (languages) => {
...
-    const decimalCharacters = getDecimalCharacters(language);
+    const locales = Intl.getCanonicalLocales(languages);
+    const formatter = Intl.NumberFormat(locales);
+    const number = -9.012345678;
 
-    characterPattern = new RegExp(
-        `[-0-9${Array.from(decimalCharacters).join('')}]`
-    );
-    characterPatternsByLanguage.set(language, characterPattern);
+    validCharacters = new Set(`${number}${formatter.format(number)}`.split(''));
```
lognaturel, line 48: Initially I thought there might be a change in behavior because this was not using `decimalCharacters`. But now I see both `decimalCharacters` and this include the `.` (here because `${number}` is included) and only one other decimal marker. Maybe there's an opportunity to reconcile those two ways of determining the set of allowed decimal markers? Doesn't feel like a big deal but it was the one thing that made me have to think through this change.

lognaturel (approved):

PR #976, opened 2023-05-15
Optimizations: do less redundant work, part 2

eyelidlessness:
While there are a variety of changes in this PR, they all follow a common theme (as the title suggests): do less redundant work, specifically DOM work. All of the changes target code which:

- contribute significantly to many forms with clear optimization potential, as identified profiling those forms
- are either static in nature, or have clearly generalizable dynamic patterns
- have reasonably clear potential for caching, or can be statically identified as no-ops for a given form

The improvements are particularly pronounced on for forms with complex behavior around repeats. But highly complex forms **without** repeats may also benefit significantly from the caching portions of the changes.

## Highlights

The commit notes hopefully cover some of the more mundane and smaller changes, but I do want to go into more detail about some new concepts and abstractions this PR introduce.

### Preprocessing model references

In this change, before a form is initialized, the view DOM is queried to identify certain elements which reference model nodesets, and to annotate those elements and some of their containers with classes which allow simpler queries during init and form entry.

In some cases this allowed certain dynamic DOM queries to be made static. This is itself a common theme across many of Enketo Core's performance challenges. While querying the DOM _in general_ can perform very well, performance degrades significantly when the queries themselves are dynamically composed, particularly when this is done repeatedly.

### Identifying form-wide characteristics

After the aforementioned preprocessing, a form's view DOM is then analyzed to determine whether certain types of functionality are ever used by the form. This analysis is then used to short-circuit whole portions of app behavior which are known to be no-ops but have a significant performance impact nonetheless.

### Static and semi-static DOM collections

This change introduces a new abstraction/class, `RefIndexedDOMCollection` (I am not married to this name! bikeshed away), which generalizes some functionality around form features which:

- Have an identifiable role/type
- Are transformed by Enketo Transformer to have a known attribute name whose value is a stable model nodeset reference, where existing form logic is concerned with:
  - finding elements of a given role by a given reference
  - identifying the presence of said reference in a form, or in another element's hierarchy
- Are indexed by their [containing] repeat instance on forms with repeats, where existing form logic is concerned with:
  - determining their current index position
  - finding an element with a given role/reference at a specified index position

In the current set of changes, the following roles/types use some or all of the above functionality:

- actions and other model references (preprocessed as described above), as well as some of their containers
- groups
- repeats and "repeatInfos" (footer elements for each run of repeat instances, which also display the add button for user-controlled repeats and may contain additional metadata or templates used by individual repeat instances)

This abstraction accounts for a very large portion of the performance improvements herein, in no small part to specific implementation details:

- Again, prior to other existing form initializtion logic, the view DOM is annotated to optimize the lookup scenarios described above. But instead of classes, comment nodes are added either inside or adjacent to the target elements. These comment nodes can be quite efficiently found using a `TreeWalker` with a `NodeFilter.SHOW_COMMENT` filter. (This same approach is used for a more minor optimization: evaluating XPath expressions against potentially empty repeat sets in the model.)

- For the duration of form init and entry, collections lazily build up caches into mappings:

  - Model reference -> elements with that reference, of the collection's role/type
  - Element -> index position

- In forms with repeats, adding or removing a repeat instance invalidates these caches, partially by reference hierarchy where possible. In forms without repeats, the caches are marked static on init and never invalidated.

There's room to optimize the caching logic further, but the improvement was already significant enough that I erred in favor of a simpler, aggressive (safer) cache invalidation scheme.

It's also possible that `TreeWalker` is not the ideal API for this set of optimiztions. I've explored other approaches, but they all come with tradeoffs and some of them degrade disproportionately for certain forms. But I'm happy to discuss other approaches if the `TreeWalker` solution feels questionable.

### The rest

The remaining functionality is pretty well described by individual commit notes, but again they generally follow a common theme of reducing redundant (or unnecessary) DOM interactions e.g. by:

- adding caching where known safe
- deferring operations which aren't needed (or needed yet)
- front loading some work to make future repetitive work less necessary
- short circuiting operations redundantly triggered at multiple depths in a DOM subtree

By relying heavily on profiling real forms of various shapes, the combined set of changes address many of the heaviest operations used by those forms. With few exceptions, the heaaviest part of form init will now be XPath evaluation (which too can be optimized, but which is also a more intrinsic expense for the domain).

One notable exception is the bench6 form's most expensive operation: dispatch of several jQuery events which are not in use by Enketo at all, almost entirely by the `bootstrap-datepicker` library, which account for ~10% of its optimized init time.

## Impact 🎉

These times were taken with [enketo-core-performance-monitor](https://github.com/enketo/enketo-core-performance-monitor), against the forms we currently monitor plus the SOAR form which has also been an optimization target. Minor modifications were made to the benchmark logic to:

- Improve accuracy by ensuring timing blocks until initial loading completes
- Improve accuracy by performing measurements directly in the headless browser to eliminate any overhead from Puppeteer
- Enable the `excludeNonRelevant` setting, because it's the generally expected/preferred behavior, and because it's also historically caused a performance penalty

I think we may want to bring these changes into the performance monitor repo.

| Form                 | Render Before | Render After    | Validate Before | Validate After  |
|----------------------|---------------|-----------------|-----------------|-----------------|
| bench6               | 4438 ms       | 1067 ms (24.0%) | 3195 ms         | 422 ms (13.20%) |
| SOAR_facilitysurvey… | 2547 ms       | 761 ms (29.8%)  | 615 ms          | 181 ms (29.43%) |
| bench1               | 1022 ms       | 648 ms (63.4%)  | 176 ms          | 47 ms (26.70%)  |
| burundi              | 782 ms        | 700 ms (89.5%)  | 355 ms          | 262 ms (73.80%) |
| calc_n_footprint_29  | 752 ms        | 483 ms (64.2%)  | 34 ms           | 16 ms (47.05%)  |
| uganda               | 714 ms        | 576 ms (80.6%)  | 400 ms          | 319 ms (79.75%) |
| bench12              | 711 ms        | 410 ms (57.6%)  | 510 ms          | 195 ms (38.23%) |
| bench9               | 699 ms        | 369 ms (52.7%)  | 24 ms           | 10 ms (41.66%)  |
| shop                 | 534 ms        | 303 ms (56.7%)  | 85 ms           | 39 ms (45.88%)  |
| bench11              | 488 ms        | 307 ms (62.9%)  | 378 ms          | 128 ms (33.86%) |
| lagtest50            | 372 ms        | 323 ms (86.8%)  | 57 ms           | 52 ms (91.22%)  |
| ukraine              | 323 ms        | 275 ms (85.1%)  | 90 ms           | 48 ms (53.33%)  |
| bench10              | 291 ms        | 253 ms (86.9%)  | 65 ms           | 47 ms (72.30%)  |
| sdiprofile           | 290 ms        | 205 ms (70.6%)  | 80 ms           | 47 ms (58.75%)  |
| haiti                | 269 ms        | 263 ms (97.7%)  | 91 ms           | 68 ms (74.72%)  |
| iraq                 | 247 ms        | 217 ms (87.8%)  | 79 ms           | 45 ms (56.96%)  |
| turkey               | 230 ms        | 158 ms (68.6%)  | 49 ms           | 29 ms (59.18%)  |
| drc                  | 224 ms        | 226 ms (100.8%) | 65 ms           | 62 ms (95.38%)  |
| bench5               | 204 ms        | 183 ms (89.7%)  | 68 ms           | 34 ms (50%)     |
| widgetsdev           | 179 ms        | 159 ms (88.8%)  | 5 ms            | 6 ms (120%)     |
| car                  | 163 ms        | 138 ms (84.6%)  | 26 ms           | 25 ms (96.15%)  |
| widgets              | 159 ms        | 133 ms (83.6%)  | 2 ms            | 4 ms (200%)     |
| bench4               | 152 ms        | 122 ms (80.2%)  | 72 ms           | 33 ms (45.83%)  |
| bench2               | 124 ms        | 114 ms (91.9%)  | 31 ms           | 14 ms (45.16%)  |
| bench3               | 116 ms        | 106 ms (91.3%)  | 10 ms           | 10 ms (100%)    |
| bench8               | 107 ms        | 94 ms (87.8%)   | 3 ms            | 4 ms (133.33%)  |
| bench7               | 107 ms        | 80 ms (74.7%)   | 19 ms           | 15 ms (78.94%)  |

### DIFF
--- a/src/js/output.js
+++ b/src/js/output.js
@@ -42,11 +42,11 @@ export default {
         let val = '';
         const that = this;
 
-        const $nodes = this.form.getRelatedNodes(
-            'data-value',
-            '.or-output',
-            updated
-        );
+        const { rootNode } = updated ?? {};
+        const $nodes =
+            rootNode == null
+                ? this.form.getRelatedNodes('data-value', '.or-output', updated)
+                : $(rootNode.querySelectorAll('.or-output'));
 
         const clonedRepeatsPresent = this.form.features.repeatClone;
 
--- a/src/js/output.js
+++ b/src/js/output.js
@@ -48,8 +48,6 @@ export default {
                 ? this.form.getRelatedNodes('data-value', '.or-output', updated)
                 : $(rootNode.querySelectorAll('.or-output'));
 
-        const clonedRepeatsPresent = this.form.features.repeatClone;
-
         $nodes.each(function () {
             const $output = $(this);
             const output = this;
@@ -92,16 +90,8 @@ export default {
                 contextPath = null;
             }
 
-            const insideRepeat =
-                clonedRepeatsPresent &&
-                $output.parentsUntil('.or', '.or-repeat').length > 0;
-            const insideRepeatClone =
-                insideRepeat &&
-                $output.parentsUntil('.or', '.or-repeat.clone').length > 0;
-            const index =
-                insideRepeatClone && contextPath
-                    ? that.form.input.getIndex(context)
-                    : 0;
+            const insideRepeat = output.closest('.or-repeat') != null;
+            const index = contextPath ? that.form.input.getIndex(context) : 0;
 
             if (typeof outputCache[expr] !== 'undefined') {
                 val = outputCache[expr];
--- a/src/js/output.js
+++ b/src/js/output.js
@@ -4,7 +4,34 @@
 
 import $ from 'jquery';
 
+/**
+ * @typedef {import('./form').Form} Form
+ */
+
 export default {
+    /**
+     * @type {Form}
+     */
+    // @ts-expect-error - this will be populated during form init, but assigning
+    // its type here improves intellisense.
+    form: null,
+
+    init() {
+        if (!this.form) {
+            throw new Error(
+                'Output module not correctly instantiated with form property.'
+            );
+        }
+
+        if (!this.form.features.output) {
+            this.update = () => {};
+
+            return;
+        }
+
+        this.update();
+    },
+
     /**
      * Updates output values, optionally filtered by those values that contain a changed node name
      *
@@ -15,21 +42,13 @@ export default {
         let val = '';
         const that = this;
 
-        if (!this.form) {
-            throw new Error(
-                'Output module not correctly instantiated with form property.'
-            );
-        }
-
         const $nodes = this.form.getRelatedNodes(
             'data-value',
             '.or-output',
             updated
         );
 
-        const clonedRepeatsPresent =
-            this.form.repeatsPresent &&
-            this.form.view.html.querySelector('.or-repeat.clone');
+        const clonedRepeatsPresent = this.form.features.repeatClone;
 
         $nodes.each(function () {
             const $output = $(this);
--- a/src/js/calculate.js
+++ b/src/js/calculate.js
@@ -19,6 +19,9 @@ export default {
     // its type here improves intellisense.
     form: null,
 
+    /** @type {WeakMap<HTMLElement, boolean>} */
+    preInitRelevance: new WeakMap(),
+
     init() {
         if (!this.form) {
             throw new Error(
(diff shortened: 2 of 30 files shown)


src/js/output.js
```diff
@@ -42,11 +42,11 @@ export default {
...
-        const $nodes = this.form.getRelatedNodes(
-            'data-value',
-            '.or-output',
-            updated
-        );
+        const { rootNode } = updated ?? {};
+        const $nodes =
+            rootNode == null
+                ? this.form.getRelatedNodes('data-value', '.or-output', updated)
+                : $(rootNode.querySelectorAll('.or-output'));
```
lognaturel, line 49: Not feeling so risky -- outputs are terminal so nothing cascades from them.

lognaturel (approved):

lognaturel:
@eyelidlessness and I did some interactive review. I have thought through each class of change and have a high-level understanding of each.

We are going to merge and update in Express master so it gets exercised on https://getodk.org/xlsform/. We'll also do a QA pass. Feedback and testing by other users is very welcome.

Issue #979, opened 2023-06-02
Unable to prefill time fields (in version 7.x)


JGreenlee:
>**Describe the bug**

When initializing a form with an `instanceStr` that has time fields filled in, the fields show "12:00 AM" regardless of what value was passed in.

> **To Reproduce**

Create a `new Form()` with an `instanceStr` in `data`, which has some time inputs with filled in values (for example, a previous response to that survey).

> **Expected behavior**

The time inputs will initialize to the values that were given in the `instanceStr`
The expected behavior is found in any Enketo version before 7.0.0.

<table>
<tr>
<td>

**using enketo-core 7.x**

<img width="368" alt="image" src="https://github.com/enketo/enketo-core/assets/15843932/dd3e383a-9a67-46b1-a246-e265f2458bf8">
</td>
<td>

**using enketo-core < 7.0.0**

<img width="363" alt="image" src="https://github.com/enketo/enketo-core/assets/15843932/2778b323-d91d-4b8b-8820-62a1934b741f">
</td>
</tr>
</table>

>**Browser and OS**

Safari on iOS. Not yet tested on other browsers.

JGreenlee:
Checking the diff between versions `6.1.7` and `7.0.0`, I didn't any changes specific to the timepicker. Maybe it's a broader issue?

Is there any reason the time format would have changed 6.x to 7.x?

JGreenlee:
I found the change that caused this regression.
https://github.com/enketo/enketo-core/commit/3accdd849662921b73966d7f29e3ffb109ddab56#diff-caad802f244c59940e6441fa0684b96b8c791470596c23de562f204c2dd1da40

It look like when dropping support for IE, this:

```js
// For IE11, we also need to strip the Left-to-Right marks \u200E...
const ds = `${new Date()
    .toLocaleDateString('en', {
        month: 'short',
        day: 'numeric',
        year: 'numeric',
    })
    .replace(/\u200E/g, '')} ${value.replace(
    /(\d\d:\d\d:\d\d)(\.\d{1,3})(\s?((\+|-)\d\d))(:)?(\d\d)?/,
    '$1 GMT$3$7'
)}`;
const d = new Date(ds);
```
was simplified to this:
```js
const ds = `${new Date().toLocaleDateString('en', {
    month: 'short',
    day: 'numeric',
    year: 'numeric',
})}`;
const d = new Date(ds);
```

[...]

PR #980, opened 2023-06-03
Fix: Respect existing val of type='time'

JGreenlee:
Fixes https://github.com/enketo/enketo-core/issues/979.

Regression and fix described there, https://github.com/enketo/enketo-core/issues/979#issuecomment-1574444845:

> I found the change that caused this regression.
> https://github.com/enketo/enketo-core/commit/3accdd849662921b73966d7f29e3ffb109ddab56#diff-caad802f244c59940e6441fa0684b96b8c791470596c23de562f204c2dd1da40
> 
> It look like when dropping support for IE, this:
> 
> ```js
> // For IE11, we also need to strip the Left-to-Right marks \u200E...
> const ds = `${new Date()
>     .toLocaleDateString('en', {
>         month: 'short',
>         day: 'numeric',
>         year: 'numeric',
>     })
>     .replace(/\u200E/g, '')} ${value.replace(
>     /(\d\d:\d\d:\d\d)(\.\d{1,3})(\s?((\+|-)\d\d))(:)?(\d\d)?/,
>     '$1 GMT$3$7'
> )}`;
> const d = new Date(ds);
> ```
> was simplified to this:
> ```js
> const ds = `${new Date().toLocaleDateString('en', {
>     month: 'short',
>     day: 'numeric',
>     year: 'numeric',
> })}`;
> const d = new Date(ds);
> ```
> 
> The intention with that commit was to remove `.replace(/\u200E/g, '')`, as IE support is no longer needed.
> I think that `${value.replace(...)` was removed by accident.
> 
> <hr>
> 
> Here is my proposed fix, which makes it a bit clearer what is actually supposed to happen here.
> ```js
> const todayDate = new Date().toLocaleDateString('en', {
>     month: 'short',
>     day: 'numeric',
>     year: 'numeric',
> });
> const valueTime = value.replace(
>     /(\d\d:\d\d:\d\d)(\.\d{1,3})(\s?((\+|-)\d\d))(:)?(\d\d)?/,
>     '$1 GMT$3$7'
> );
> const d = new Date(`${todayDate} ${valueTime}`);
> ```
> 
> I will put together a PR for this.

### DIFF
--- a/src/js/input.js
+++ b/src/js/input.js
@@ -326,12 +326,16 @@ export default {
                     // Use today's date to incorporate daylight savings changes,
                     // Strip the thousands of a second, because most browsers fail to parse such a time.
                     // Add a space before the timezone offset to satisfy some browsers.
-                    const ds = `${new Date().toLocaleDateString('en', {
+                    const todayDate = new Date().toLocaleDateString('en', {
                         month: 'short',
                         day: 'numeric',
                         year: 'numeric',
-                    })}`;
-                    const d = new Date(ds);
+                    });
+                    const valueTime = value.replace(
+                        /(\d\d:\d\d:\d\d)(\.\d{1,3})(\s?((\+|-)\d\d))(:)?(\d\d)?/,
+                        '$1 GMT$3$7'
+                    );
+                    const d = new Date(`${todayDate} ${valueTime}`);
                     if (d.toString() !== 'Invalid Date') {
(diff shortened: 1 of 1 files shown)


eyelidlessness (approved):
Thanks!

lognaturel:
The same raster test is failing as fixed in the last merged PR. So I think we can merge this and it should be green on master, right?

PR #983, opened 2023-06-28
Fix: TOC navigation in pages theme

eyelidlessness:
Fixes enketo/enketo-express#568. This was a regression introduced by an overzealous find/replace in #969.


lognaturel:
> have also begun to fail after apparent changes in Chromium's rasterization

Ugh!

lognaturel (approved):

PR #985, opened 2023-07-11
use map instead of object

delcroip:
it seems that maps are faster, as this is assessed many time small speed improvement can have good effect

https://www.zhenghao.io/posts/object-vs-map

I cannot test (no set up working) but this should not require too many change to work

### DIFF
--- a/src/js/form-model.js
+++ b/src/js/form-model.js
@@ -1424,7 +1424,8 @@ FormModel.prototype.evaluate = function (
     cacheable = original === expr;
 
     // if no cached conversion exists
-    if (!this.convertedExpressions.has(cacheKey)) {
+    cached_expr = this.convertedExpressions.get(cacheKey)
+    if (cached_expr === undefined) {
         expr = expr.trim();
         expr = this.replaceInstanceFn(expr);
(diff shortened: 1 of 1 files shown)


eyelidlessness (approved):
Thanks!

I don't expect this to have a significant performance impact in itself (though more broadly adopting `Map` may, in aggregate!). Even apart from performance, though, I've generally been wanting to move towards using:

- `Map` for "bag of stuff" key-value data types
- objects for data of a statically known type/shape (e.g. with explicitly known keys)

This is a good first step in that directions, and it's a good opportunity to put the intent into words.

The only thing that felt potentially risky about this change in particular is the possibility that `FormModel#convertedExpressions` may be accessed elsewhere, but I do not see any such access.

I want to more thoroughly evaluate where we're using instance properties this way, and whether they can be either converted to local (e.g. to module or other more narrowly scoped) variables, better encapsulated or at least documented as private. This is theoretically a "breaking" change, in that it modifies an _accessible_ API type, but I don't think that technicality applies for real world usage in this case. I just want to make note that this was a consideration in review.

PR #994, opened 2023-08-15
Fix: global initialization of widgets where repeat count is zero

eyelidlessness:
Fixes enketo/enketo-express#580.


lognaturel (approved):
How unexpected and frustrating that there was no test that started from a blank count.

PR #995, opened 2023-08-16
Fix: multiple "static" itemset datalists in the same repeat

eyelidlessness:
This fixes a case identified by the [recent pyxform change](https://github.com/XLSForm/pyxform/pull/614) to always generate secondary instances for selects. Hopefully the in-code commentary is enough elaboration on the nature of the bug and its fix.

### DIFF
--- a/test/forms/repeat-multiple-shared-datalist-itemsets.xml
+++ b/test/forms/repeat-multiple-shared-datalist-itemsets.xml
@@ -1,0 +1,165 @@
+<?xml version="1.0"?>
+<!--
+    This fixture is used in tests for an itemset/repeat bug. Briefly, forms with
+    the following conditions fail to load:
+
+    - The form contains a repeat
+    - The repeat contains at least two itemsets which:
+        - Have a "static" (details and commentary inline below) reference to
+          items in a secondary instance
+        - Are presented as a combination of `input[list]` and `datalist`
+          elements
+
+    Much of the inline commentary below is intended to:
+
+    - clarify implementation trivia, explaining the nature of this bug and its
+      various contributing factors
+
+    - consolidate the explanation into a single place, as the various
+      contributing factors are spread across multiple files/modules, across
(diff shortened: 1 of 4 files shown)


test/forms/repeat-multiple-shared-datalist-itemsets.xml
```diff
@@ -0,0 +1,165 @@
...
+    Much of the inline commentary below is intended to:
+
+    - clarify implementation trivia, explaining the nature of this bug and its
+      various contributing factors
+
+    - consolidate the explanation into a single place, as the various
+      contributing factors are spread across multiple files/modules, across
+      multiple projects
+
+    Such information is probably better suited for, perhaps, a knowledge base
```
lognaturel, line 22: Have you seen this done in an effective way? From my experience, inline comments are about as good as it gets, especially with keeping things up to date. The ideal is of course not to have such unexpected interactions so that commentary isn't needed at all but for now this seems like the best spot so this paragraph could be removed.

lognaturel (approved):
I have verified that this also works with the user-provided full form. 

@eyelidlessness I'll let you decide whether you want to cut out the paragraph about alternative places for the commentary.

test/forms/repeat-multiple-shared-datalist-itemsets.xml
eyelidlessness: I've seen effective use of internal documentation tools for e.g. cross-cutting concerns **where sufficiently embraced**. I'm not sure how effective it'd be with the current structure or how we'd best utilize it, but it's worth thinking about. For now, I'm fine with removing the paragraph.

I will say I'm concerned that much of this commentary well be more valuable upfront than over time, as its connection to the behaviors it describes links *to* the pertinent code rather than from it. But I'm not sure how we'd best solve that either.

lognaturel (approved):

PR #998, opened 2023-09-02
Update config.js use correct tile.osm.org URL

Firefishy:
See: https://github.com/openstreetmap/operations/issues/737


lognaturel:
Thanks very much! We're in process of migrating to a monorepo and I've opened an equivalent PR there: https://github.com/enketo/enketo/pull/11

